Issue #2, opened 2015-06-11
Invalid baudrate error handling


rsluiter:
When asyn record posts this error message;
"rls2:asyn_9: Error setting option, SIO_BAUD_SET failed: S_errno_EINVAL"
it does not revert the baudrate in either the BAUD or LBAUD fields to a valid value.
Investigate if asyn record can revert the BAUD and/or LBAUD field(s) to the previous, valid value after detecting this error (and also learn how to use github).

MarkRivers:
Fixed.  Did not use Ron's pull request because it is not necessary to change the asynRecord at all.  It is sufficient to have the driver restore the previous value of the baud rate when setting it fails.  Same for all termios settings.

PR #3, opened 2015-06-12
fix for issue #2

rsluiter:
Made source code modifications to make asyn record revert to the previous valid baudrate after an invalid value is requested.
This fixes issue #2

### DIFF
--- a/asyn/asynRecord/asynRecord.c
+++ b/asyn/asynRecord/asynRecord.c
@@ -378,6 +378,10 @@ static long special(struct dbAddr * paddr, int after)
     asynQueuePriority priority;
 
     if(!after) {
+        if (fieldIndex == asynRecordBAUD)   /* Save previous value in case requested value is invalid. */
+            pasynRec->pbaud = pasynRec->baud;
+        else if (fieldIndex == asynRecordLBAUD)
+            pasynRec->plbaud = pasynRec->lbaud;
         return 0;
     }
     resetError(pasynRec);
@@ -1766,11 +1770,15 @@ static void setOption(asynUser * pasynUser)
     case asynRecordBAUD:
         status = pasynRecPvt->pasynOption->setOption(pasynRecPvt->asynOptionPvt,
             pasynUser, "baud", baud_choices[pasynRec->baud]);
+        if (status != asynSuccess)
+            pasynRec->baud = pasynRec->pbaud;
         break;
     case asynRecordLBAUD:
         sprintf(optionString, "%d", pasynRec->lbaud);
         status = pasynRecPvt->pasynOption->setOption(pasynRecPvt->asynOptionPvt,
             pasynUser, "baud", optionString);
+        if (status != asynSuccess)
+            pasynRec->lbaud = pasynRec->plbaud;
         break;
     case asynRecordPRTY:
         status = pasynRecPvt->pasynOption->setOption(pasynRecPvt->asynOptionPvt,
--- a/asyn/asynRecord/asynRecord.dbd
+++ b/asyn/asynRecord/asynRecord.dbd
@@ -365,11 +365,22 @@ recordtype(asyn) {
         special(SPC_MOD)
         menu(serialBAUD)
     }
+    field(PBAUD,DBF_MENU) {
+        prompt("Last valid BAUD")
+        interest(2)
+        special(SPC_NOMOD)
+        menu(serialBAUD)
+    }
     field(LBAUD,DBF_LONG) {
         prompt("Baud rate")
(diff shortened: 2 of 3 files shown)


MarkRivers:
I have a comment on the proposed changes to drvAsynSerialPort.c.

In setOption whenever applyOptions returns an error the baud rate then tty->baud is set back to the previous baud rate.  But applyOptions is called not only when the baud rate is changed, but also whenever any other option is changed.  Why is baud rate being treated specially?  What if ixoff causes an error, shouldn't it be set back to its original value as well?  Should the entire tty structure be saved and then restored if there is an error? 

These same comments go for the asynRecord as well.  Why is the baud rate being treated differently from other settings that might also return an error?

MarkRivers:
Ron, I have not heard back from you about my July 14 comment on this pull request.

rsluiter:
On a VxWorks IOC, your specific example of ixoff (and ixany) are handled at a high level;
"Error setting option, Option ixoff not supported on vxWorks"
"Error setting option, Option ixany not supported on vxWorks"
without any call to the low level tty ioctrl function.

Currently there is no error status returned from either tyGSOctalOptsSet nor ip520OptsSet. Only the baudrate set functions (tyGSOctalOptsSet and IP520BaudSet) return an error.

That said, you are correct that all the terminal characteristics should be processed the same, but on a VxWorks platform that extra work will not yield results.

MarkRivers:
I have come up with a much simpler solution.  The changes you made to asynRecord.c and asynRecord.dbd are not necessary.  It is sufficient to simply restore the previous values of tty->baud and tty->termios when dvAsynSerialPort.c::applyOptions() returns an error.  The reason is that asynRecord always calls getOptions() immediately after setting any option, so it will then read back the actual value of the option from the driver.  I have tested with vxWorks, setting baud to an invalid value.  It immediately restores back to the previous correct value.  I have created a new branch, serial_port_fix, and issued a new pull request for it.  Please take a look and test it.

MarkRivers:
Folks,

I made a mistake on the serial_port_fix branch to accidentally committed directly to master.

This is the change I made:

***********
corvette:asyn/asyn/drvAsynSerial>git diff HEAD^
diff --git a/asyn/drvAsynSerial/drvAsynSerialPort.c b/asyn/drvAsynSerial/drvAsynSerialPort.c
index 090bbbb..7efb367 100644
--- a/asyn/drvAsynSerial/drvAsynSerialPort.c
+++ b/asyn/drvAsynSerial/drvAsynSerialPort.c
@@ -210,10 +210,17 @@ static asynStatus
setOption(void *drvPvt, asynUser *pasynUser, const char *key, const char *val)
{
     ttyController_t *tty = (ttyController_t *)drvPvt;
+    struct termios termiosPrev;
+    int baudPrev;
     assert(tty);
     asynPrint(pasynUser, ASYN_TRACE_FLOW,
                     "%s setOption key %s val %s\n", tty->portName, key, val);
+
+    /* Make a copy of tty->termios and tty->baud so we can restore them in case of errors */
+    termiosPrev = tty->termios;
+    baudPrev = tty->baud;
+
     if (epicsStrCaseCmp(key, "baud") == 0) {
         int baud;
         if(sscanf(val, "%d", &baud) != 1) {
@@ -464,8 +471,14 @@ setOption(void *drvPvt, asynUser *pasynUser, const char *key, const char *val)
                                                 "Unsupported key \"%s\"", key);
         return asynError;
     }
-    if (tty->fd >= 0)
-        return applyOptions(pasynUser, tty);
+    if (tty->fd >= 0) {
+        if (applyOptions(pasynUser, tty) != asynSuccess) {
+            /* Restore previous values of tty->baud and tty->termios */
+            tty->baud = baudPrev;
+            tty->termios = termiosPrev;
+            return asynError;
+        }
+    }
     return asynSuccess;
}
static const struct asynOption asynOptionMethods = { setOption, getOption };
***********

So setOption now saves the previous values of tty->baud and tty->termios at the beginning, and restores those values if applyOptions() returns an error.

Does this look OK?

I have tested it with the asynRecord on VxWorks.  Previously if I selected an invalid baud rate with the asyn record then it would continue to show that invalid baud rate even after the driver printed an error message.  Now it reverts to the previous valid baud rate.  The reason that it now works is that in the asynRecord after an option is changed it always calls getOptions() to read the actual option values back from the driver. Previously drvAsynSerialPort was returning the invalid value that had been passed but failed.  Now it reverts to the previous (valid) value when applyOptions returns an error.

Mark

rsluiter:
I tested these changes on VxWorks and they look good; the baudrate and 
termios(?) are restored after an error.

PR #4, opened 2015-06-21
Allow COMxx syntax for higher numbered Windows serial ports

FreddieAkeroyd:
Serial port names such as COM1 to COM9 are reserved words, but devices
such as COM18 normally need to be referenced in device namespace via \\.\COM18
This commit allows serial ports such as COM18 to be referenced unprefixed


MarkRivers:
Freddie, I'm assuming you have tested this on both low numbered COM ports and high numbered COM ports?

FreddieAkeroyd:
Hi Mark, yes I've tested the code with both high and low numbered COM ports. Regards, Freddie

PR #5, opened 2015-07-01
drvAsynIPPort: Add USERFLAG_CLOSE_ON_READ_TIMEOUT

tboegi:
Different EPICS records may share one TCP/IP connection (like different
axes at the same motion controller unit).

When the cable is pulled, TCP will buffer data send buffer,
and the readIt() function will return a timeout.

The default timeout is 2.0 seconds, and after this timeout the first axis
is put into the alarm state.

The next axis will need another 2 seconds on it's own, and for an 8-axes
system all axes are in alarm state after 16 seconds.

On the other hand, all the poll() requests are still in the send buffer,
so when the cable is put back, all 8 poll() requests are send to the
controller at once.

This may be a non-ideal situation: The controller must process many
messages at once, and a bunch of responses is send to the IOC.
The IOC should discard them all in flushIt(), or they may end up on the wrong
axis, depending on the timing.

Make it possible to close() the TCP connection already at the first timeout.
This means that all axes go into the alarm state at the same time,
and all data coming back from the controller are reliable discarded in the
TCP layer.

The new feature can be activated like this:
drvAsynIPPortConfigure2("MC_CPU1","192.168.1.15:5024",0,0,0,1)

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPPort.c
@@ -102,6 +102,7 @@ typedef struct {
     char              *portName;
     int                socketType;
     int                flags;
+    int                userFlags;
     SOCKET             fd;
     unsigned long      nRead;
     unsigned long      nWritten;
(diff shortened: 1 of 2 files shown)


tboegi:
Will re-submit as suggested by Eric

PR #6, opened 2015-08-18
drvAsynIPPort: configurable disconnect on read timeout

tboegi:
Different EPICS records may share one TCP/IP connection (like different
axes at the same motion controller unit).

When the cable is pulled, TCP will buffer data send buffer,
and the readIt() function will return a timeout.

The default timeout is 2.0 seconds, and after this timeout the first axis
is put into the alarm state.

The next axis will need another 2 seconds on it's own, and for an 8-axes
system all axes are in alarm state after 16 seconds.

On the other hand, all the poll() requests are still in the send buffer,
so when the cable is put back, all 8 poll() requests are send to the
controller at once.

This may be a non-ideal situation: The controller must process many
messages at once, and a bunch of responses is send to the IOC.
The IOC should discard them all in flushIt(), or they may end up
on the wrong axis, depending on the timing.

Make it possible to close() the TCP connection already at the first timeout.
This means that all axes go into the alarm state at the same time,
and all data coming back from the controller are reliable discarded in the
TCP layer.

The new feature can be activated like this:
drvAsynIPPortConfigure("MC_CPU1","192.168.1.15:5024",0,2,0)

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPPort.c
@@ -102,6 +102,7 @@ typedef struct {
     char              *portName;
     int                socketType;
     int                flags;
+    int                userFlags;
     SOCKET             fd;
     unsigned long      nRead;
     unsigned long      nWritten;
@@ -123,6 +124,11 @@ typedef struct {
 #define FLAG_SHUTDOWN                   0x4
 #define FLAG_NEED_LOOKUP                0x100
 #define FLAG_DONE_LOOKUP                0x200
+
+#define USERFLAG_NO_PROCESS_EOS         0x1
+#define USERFLAG_CLOSE_ON_READ_TIMEOUT  0x2
+#define USERFLAG_RESERVED               (~0x3)
+
 
 #ifdef FAKE_POLL
 /*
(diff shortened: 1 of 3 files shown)


asyn/drvAsynSerial/drvAsynIPPort.c
```diff
@@ -693,6 +699,7 @@ drvAsynIPPortConfigure(const char *portName,
     asynStatus status;
     char *cp;
     int nbytes;
+    int autoConnect = noAutoConnect & USERFLAG_NO_AUTO_CONNECT ? 0 : 1;
```
anjohnson: I would suggest renaming the `int noAutoConnect` argument to the `drvAsynIPPortConfigure()` function (and change its name in the documentation too) to avoid future confusion. The obvious name would be `userFlags`. You should probably also check that the value doesn't have any unused bits set — if the code doesn't object to them now, it'll break IOC applications that might set them by mistake or start using them for other things when the next bit in userFlags gets allocated. Also don't use a conditional operator when the logical meaning is 'not':
``` diff
-int autoConnect = noAutoConnect & USERFLAG_NO_AUTO_CONNECT ? 0 : 1;
+int autoConnect = ! (userFlags & USERFLAG_NO_AUTO_CONNECT);
```

MarkRivers:
I don't like adding another bit to the noAutoConnect flag.  That flag is handled by asynManager (after being negated).  We will be combining bits that are handled by asynManager (noAutoConnect) with bits that are specific to the drvAsynIPPort driver.  What if asynManager needs to add additional bits to the autoConnect flag to pasynManager->registerPort() in the future?  We would have a real mess, because those bit locations would be different from the ones used by drvAsynIPPort.  I suggest that it is cleaner to just add a new final disconnectOnTimeout argument to the drvAsynIPPortConfigure command:

`epicsShareFunc int drvAsynIPPortConfigure(const char *portName,
                                          const char *hostInfo,
                                          unsigned int priority,
                                          int noAutoConnect,
                                          int noProcessEos,
                                          int disconnectOnTimeout);`

disconnectOnTimeout=0 is the current behavior.  This is backwards compatible in the vxWorks and iocsh shells, since missing arguments default to 0.  If someone is calling drvAsynIPPortConfigure from C or C++ they will get a compiler error, which will flag that they need to change their code.

MarkRivers:
I just looked at your original pull request #5 and it is very similar to what I am suggesting.  However, you added a new iocsh command (drvAsynIPPOrtConfigure2) which I don't think is necessary.

Alternatively if we don't want to add another argument to the configuration command then we could rename the noProcessEos flag to be userFlags, and add the disconnectOnTimeout bit there.  That would be OK with me, because the noProcessEos argument is specific to the drvAsynIPPort driver, unlike noAutoConnect which is handled by asynManager.

PR #9, opened 2015-10-08
Add ASYN_TRACE_INFO

tboegi:
In case that the traces ASYN_TRACE_FLOW or ASYN_TRACEIO_DRIVER
are too spammy, add ASYN_TRACE_INFO

This can be used to trace commands send to a device like a motor,
or selective status changes.

### DIFF
--- a/asyn/asynDriver/asynDriver.h
+++ b/asyn/asynDriver/asynDriver.h
@@ -188,6 +188,7 @@ typedef struct  asynLockPortNotify {
 #define ASYN_TRACEIO_DRIVER  0x0008
 #define ASYN_TRACE_FLOW      0x0010
 #define ASYN_TRACE_WARNING   0x0020
+#define ASYN_TRACE_INFO      0x0040
 
 /* traceIO mask definitions*/
 #define ASYN_TRACEIO_NODATA 0x0000


ulrikpedersen:
Hi @tboegi 

I think its a good idea to add another trace flag - we sometimes end up adding an extra custom trace flag like this for specific drivers anyway. However, your patch is lacking a few bits before it is really useful - mainly the bits that lets us control it through the asynRecord.

FYI a similar flag (ASYN_TRACE_WARNING) was added in June 2013. The tech-talk discussion is [here](http://www.aps.anl.gov/epics/tech-talk/2013/msg00829.php). If you have a look at the history of that series of commits you can see how the flag was added to the asynRecord. The relevant changes start from commit b605da8a1a2a3bdb88892ee6a88f70ab015e568d

tboegi:
I realize 3 problems:
1) The repo from the PR does not exist any more
2) The "INFO" was too confusing
3) I don't have any of EDM, MEDM or whatever UI tools available.

However, 1) and 2) are addressed here:
https://github.com/epics-modules/asyn/pull/63

tboegi:
See #63

PR #15, opened 2016-02-09
epicsInterruptibleSyscall.c: Use format string for cantProceed()

tboegi:


MarkRivers:
What's the motivation for this change, since they should be equivalent?

tboegi:
1st motivation is a compiler warning - under Mac OS (if that matters)

The compiler warning is motivated by the fact that msg could be something like
"%s", and then printf() may crash.

So printing a message without a format string "%s" is deprecated these days.

MarkRivers:
OK, I understand.

PR #18, opened 2016-02-13
Added asynExceptionToString function and use it in error msgs.

bhill-slac:
Also fixed some minor typos and compiler warnings.


MarkRivers:
I like this change.  However, tt looks like your changes to asynDriver.h and perhaps asynRecord.c have tab characters.  If so can you please replace them with spaces?

MarkRivers:
Never mind, I'll fix them.

PR #19, opened 2016-02-13
Added paramName to trace msgs

bhill-slac:


klauer:
This is a very useful addition - I often end up tagging on `paramName` to trace messages in asyn-based drivers I write.

PR #22, opened 2016-05-02
Add RS485 support to drvAsynSerialPort driver

ffeldbauer:
Under linux RS485 interfaces can be controlled via the kernel dirver.
For this, the struct serial_rs485 defined in linux/serial.h has to be used.

The settings for RS485 are added as asynOptions to the drvAsynSerialPort:
rs485_enable                -- Enable RS485 support for interface in kernel
rs485_rts_on_send           -- Set logic level of RTS signal while sending ('Y' = HIGH)
rs485_rts_after_send        -- Set logic level of RTS signal after sending ('Y' = HIGH)
rs485_delay_rts_before_send -- Delay before send (milliseconds)
rs485_delay_rts_after_send  -- Delay after send (milliseconds)

### DIFF
--- a/asyn/drvAsynSerial/drvAsynSerialPort.c
+++ b/asyn/drvAsynSerial/drvAsynSerialPort.c
@@ -50,6 +50,18 @@
 # include <termios.h>
 #endif
 
+/*
+ * RS485 support on linux-arm
+ */
+#ifdef linux /* RS485 support */
+# include <sys/ioctl.h>
+# include <linux/serial.h>
+
+# define TIOCGRS485      0x542E
+# define TIOCSRS485      0x542F
+#endif
+
+
 #ifdef vxWorks
 /*
  * Fake termios structure
@@ -71,6 +83,9 @@ typedef struct {
     unsigned long      nRead;
     unsigned long      nWritten;
     struct termios     termios;
+#ifdef linux /* RS485 support */
+    struct serial_rs485  rs485;
+#endif
     int                baud;
     double             readTimeout;
     double             writeTimeout;
@@ -120,6 +135,17 @@ applyOptions(asynUser *pasynUser, ttyController_t *tty)
         return asynError;
     }
 #endif
+
+#ifdef linux /* RS485 support */
+    if( tty->rs485.flags != 0 ) {
+      if( ioctl( tty->fd, TIOCSRS485, &tty->rs485 ) < 0 ) {
+        epicsSnprintf(pasynUser->errorMessage, pasynUser->errorMessageSize,
+                                   "ioctl TIOCSRS485 failed: %s", strerror(errno));
+        return asynError;
+      }
+    }
+#endif
+
     return asynSuccess;
 }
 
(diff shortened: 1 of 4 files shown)


norumwe12:
I suggest that the conditionals be changed to something like 
#idef ENABLE_RS85
or something like that.  There would be one conditional at the top, currently
#ifdef linux
# define ENABLE_RS485
#endif
This would make it easier to add this support to other platforms that might provide the same ‘struct serial_rs485’ in the future.

MarkRivers:
I was going to suggest exactly the same thing.

Thanks,
Mark


Sent from my iPhone

On May 2, 2016, at 10:52 AM, norumwe12 <notifications@github.com<mailto:notifications@github.com>> wrote:

I suggest that the conditionals be changed to something like
#idef ENABLE_RS85
or something like that. There would be one conditional at the top, currently
#ifdef linux
# define ENABLE_RS485
#endif
This would make it easier to add this support to other platforms that might provide the same �struct serial_rs485� in the future.

> On May 2, 2016, at 7:49 AM, Florian Feldbauer <notifications@github.com<mailto:notifications@github.com>> wrote:
>
> Under linux RS485 interfaces can be controlled via the kernel dirver.
> For this, the struct serial_rs485 defined in linux/serial.h has to be used.
>
>


�

norumwe12:
I thought of an alternative that I think is better.  
Remove all the conditionals and then at the top have

# define TIOCGRS485      0x542E
# define TIOCSRS485      0x542F

#ifdef linux
  /* RS485 support on linux */
# include <sys/ioctl.h>
# include <linux/serial.h>
#else 
  /* Dummy declarations RS485 code */
  struct serial_rs485 {
        int flags;
        unsigned int delay_rts_before_send;
        unsigned int delay_rts_after_send;
  };
# define SER_RS485_ENABLED        0x1
# define SER_RS485_RTS_ON_SEND    0x2
# define SER_RS485_RTS_AFTER_SEND 0x4
#endif


This suggestion brought to you by,  "the committee to limit the number of conditional tests in EPICS code”.

anjohnson:
The EPICS build system allows you to provide and install OS-specific header files, just add them to an os/<os-class> subdirectory of your src directory. This permits the complete removal of conditionals from the code, at the expense of one additional #include statement. Use the os/default directory to hold the version that should apply to OSs not otherwise provided for. Take a look at Base/src/libCom/osi/os/* for examples.

MarkRivers:
I'm not sure an OS-dependent header file is the best solution in this case, because we might want to support RS485 on other archs (e.g. vxWorks) in the future.  There are already lots of #ifdef vxWorks in the code now.  In fact so many that perhaps we should have made a separate driver file for vxWorks as we did for Windows?

anjohnson:
You can also have OS-specific source files in the same place as well, I just responded to Eric's suggestion which only contained declarations. The directory structure means you don't have to have any conditionals in your Makefile, you just let it pick the source file to build from the appropriate directory.

ralphlange:
devIocStats is another example - outside of Base.

ffeldbauer:
I guess this OS-specific header ansatz might even simplify implementing this support for other archs.

ffeldbauer:
Another question: The settings for RS485 are send to the kernel via IOCTL.
Should this also be made os-specific? Or is it safe to include "sys/ioctl.h" for all OS? (I never worked with other OS then linux...)

ralphlange:
Not for all OS, that's for sure. I would say
- It is safe to assume that vxWorks and Windows are different.
- The Linux man pages usually declare if an API is POSIX conformant. If yes -> posix, else ->Linux.

ffeldbauer:
For me, it looks like it is not POSIX conformant

> CONFORMING TO
>       No  single standard. ...

So, "sys/ioctl.h" is included in the linux-header, while the default-header has now a dummy implementation of the ioctl function.

MarkRivers:
I just downloaded this to my system.  It did not work on vxWorks.  I fixed that and pushed back to a new rs485 branch I just created.  However, it also does not build on older Linux systems, and I am not sure how to fix that.  These are the errors;

    idea:epics/devel/asyn>make -sj
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:76: error: field 'rs485' has incomplete type
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c: In function 'getOption':
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:207: error: 'SER_RS485_ENABLED' undeclared (first use in this function)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:207: error: (Each undeclared identifier is reported only once
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:207: error: for each function it appears in.)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:210: error: 'SER_RS485_RTS_ON_SEND' undeclared (first use in this function)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:213: error: 'SER_RS485_RTS_AFTER_SEND' undeclared (first use in this function)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c: In function 'setOption':
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:241: error: storage size of 'rs485Prev' isn't known
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:499: error: 'SER_RS485_ENABLED' undeclared (first use in this function)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:512: error: 'SER_RS485_RTS_ON_SEND' undeclared (first use in this function)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:525: error: 'SER_RS485_RTS_AFTER_SEND' undeclared (first use in this function)
    ../../asyn/drvAsynSerial/drvAsynSerialPort.c:241: warning: unused variable 'rs485Prev'
    make[2]: *** [drvAsynSerialPort.o] Error 1 
    make[2]: *** Waiting for unfinished jobs....  
    make[1]: *** [install.linux-x86-gcc43] Error 2
    make: *** [asyn.install] Error 2

PR #25, opened 2016-06-01
Setup asynOctet records with initial values if necessary

hdante:
If the driver returns asynSuccess when reading during the init_record
device support function, the output record should use the returned
value as the initial value. This already works for numeric interfaces.
This patch implements the same for the octet interface.

### DIFF
--- a/asyn/devEpics/devAsynOctet.c
+++ b/asyn/devEpics/devAsynOctet.c
@@ -190,6 +190,8 @@ static long initCommon(dbCommon *precord, DBLINK *plink, userCallback callback,
     asynInterface *pasynInterface;
     commonDset    *pdset = (commonDset *)precord->dset;
     asynOctet     *poctet;
+    size_t        nBytesRead;
+    int           eomReason;
 
     pdevPvt = callocMustSucceed(1,sizeof(*pdevPvt),"devAsynOctet::initCommon");
     precord->dpvt = pdevPvt;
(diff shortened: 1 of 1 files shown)


MarkRivers:
Hi Henrique,
Thanks for this.  I think it needs 2 changes:
1) The read operation must be done with pasynOctetSyncIO->read().  poctet->read can only be called from the callback function via pasynManager->queueRequest.  You will see that devAsynInt32 uses pasynInt32SyncIO->read(), devAsynFloat64 uses pasynFloat64SyncIO->read(), etc. You must first call pasynOctetSyncIO->connect(), and at the end you should called pasynOctetSyncIO->disconnect(). Note that devAsynInt32 does not call disconnect(), but devAsynFloat64 and devAsynUInt32Digital do, and there is a small memory leak if you don't.
2) You are reading directly into the record field, because that is what pValue points to.  I don't think you should do that: what if the read operation returns an error, but the driver has modified the value for some reason?  I think you need to read into a scratch buffer, and only copy to the record field if the driver returns asynSuccess.

MarkRivers:
I have made the changes I noted above and pushed to a new asynOctet_fix branch.  I have modified the testErrorsApp application to allow testing this change.  It seems to work for stringout, stringin, waveform out if the info tag asyn:READBACK is 1.  However, it is not working for waveform output records without asyn:READBACK=1, which does not seem correct.  Needs more work.

Issue #35, opened 2016-12-13
avoid static param table size?


mdavidsaver:
Is there any interest in making redundant the paramTableSize argument of the asynPortDriver ctor?  I have found the FIRST/LAST macro pattern used to calculate it is error prone.  It seems to me that it would be a simple matter to change paramList::createParam() to resize its internal storage as new parameters are added.

I am not proposing to remove the paramTableSize argument as this would break too much, just to ignore it (or maybe warn).  Only private data members should change, so existing code would continue to work as before.

MarkRivers:
That seems like a good idea.  When I wrote asynPortDriver it was my first C++ project, and I was not aware of the nice things in std:: that can make appending to lists easy, etc.

Last week I was also thinking that we should change the asynPortDriver parameter library to use std::string for string parameters, so that we don't need to pass a maximum length to getStringParam.  The existing method which passes a char * would not be changed, but would be supplemented with a method that gets an  std::string without needing to specify a maximum size.  There are limitations in paramAttribute in ADCore that come from this limitation in getStringParam.

mp49:
Any idea how using std::string or STL stuff would affect compatibility with other platforms like VxWorks? This is not so much a concern for areaDetector, but more of a concern for asynPortDriver.

mdavidsaver:
I don't think this usage will be a problem, even for vxworks 5.5 (my worst case for C++ compatibility).  In any event, APS jenkins should catch a compile failure https://jenkins.aps.anl.gov/job/epics-asyn-vx55/ and a sufficiently motivated person could run the new asynPortDriverTest to be certain (I have no access to vxworks).

MarkRivers:
I merged #36  which addresses this problem.  Note that when I tried to build it failed because of the order of libraries in asynPortDriver/unittest/Makefile.  I fixed that and pushed back to github.  I have built on linux-x86_64
linux-x86
vxWorks-ppc32 (6.9)
linux-x86.win32-x86-mingw 
linux-x86.win32-x86-mingw-static
windows-x64-static (VS 2010 and VS 2015)

MarkRivers:
[...]

PR #36, opened 2016-12-14
avoid static param table size

mdavidsaver:
As described in #35 remove static limit on parameter size.  Also create asynPortDriverTest to exercise parameter creation and change flag code effected by this change.  Remove unused (and un-buildable) ParamListTest.cpp and ParamValTest.cpp.

Also some bug fixes found along the way.  An incorrect error check in ```asynPortClient::asynPortClient```.  Freeing of otherwise leaked interruptNodes in all *Base except asynOctetBase.  Finally, make ```~asynInt32Client``` call cancelInterruptUser if necessary.  Other *Client classes will need this as well, but I'm out of time for today.


asyn/asynPortDriver/asynPortDriver.cpp
```diff
@@ -77,66 +80,64 @@ class paramList {
...
     if (this->findParam(name, index) == asynSuccess) return asynParamAlreadyExists;
-    *index = this->nextParam++;
-    if (*index < 0 || *index >= this->nVals) return asynParamBadIndex;
-    delete this->vals[*index];
-    this->vals[*index] = new paramVal(name, type);
+
+    std::auto_ptr<paramVal> param(new paramVal(name, type));
+
+    if (this->vals.size() == maxParams) {
+        asynPrint(pasynPortDriver->pasynUserSelf, ASYN_TRACE_ERROR,
```
ulrikpedersen: Use ASYN_TRACE_WARNING instead of ERROR?

asyn/asynPortDriver/asynPortDriver.cpp
```diff
@@ -146,8 +147,11 @@ asynStatus paramList::createParam(const char *name, asynParamType type, int *ind
   * \return Returns asynParamNotFound if name is not found in the parameter list. */
 asynStatus paramList::findParam(const char *name, int *index)
 {
-    for (*index=0; *index<this->nVals; (*index)++) {
-        if (this->vals[*index]->nameEquals(name)) return asynSuccess;
+    for (size_t i=0; i<this->vals.size(); i++) {
```
ulrikpedersen: vals is a vector - why not use an iterator over vals?

asyn/asynPortDriver/asynPortDriver.cpp
mdavidsaver: will do

asyn/asynPortDriver/asynPortDriver.cpp
mdavidsaver: vector is the one container where this makes no practical difference.  It is a smaller change, and easier to validate, to keep the indexed access.

PR #38, opened 2017-01-03
Change internal storage of string parameters from char * to std::string *.

MarkRivers:


MarkRivers:
Changed the internal storage of string parameters from char * to std::string.

Added new setStringParam() and getStringParam() methods that take an std::string&amp;
argument rather than char *. These do not require specifying a maximum string length, unlike the char * methods.

asyn/asynPortDriver/asynPortDriver.cpp
```diff
@@ -56,10 +56,12 @@ class paramList {
     asynStatus setUInt32(int index, epicsUInt32 value, epicsUInt32 valueMask, epicsUInt32 interruptMask);
     asynStatus setDouble(int index, double value);
     asynStatus setString(int index, const char *string);
+    asynStatus setString(int index, std::string& string);
```
mdavidsaver: setter type methods should take a const reference when the argument isn't modified (the same way the other overload takes 'const char*' instead of 'char*').

```
asynStatus setString(int index, const std::string& string);
```

asyn/asynPortDriver/paramVal.h
```diff
@@ -54,7 +56,7 @@ class paramVal {
         epicsInt32   ival;
         epicsUInt32  uival;
         epicsFloat64 dval;
-        char         *sval;
+        std::string  *sval;
```
mdavidsaver: yuck.  I'd suggest just moving ```sval``` outside the union as a non-pointer member.  This is a bit bigger, but not much.  I think this is offset by the code simplification/safety and avoiding an extra indirection.  For me (linux on x86_64 w/ gcc 4.9.2) ```sizeof(std::string)==8```.

asyn/asynPortDriver/paramVal.h
```diff
@@ -31,8 +33,8 @@ class paramVal {
     epicsUInt32 getUInt32(epicsUInt32 valueMask);
     void setDouble(epicsFloat64 value);
     epicsFloat64 getDouble();
-    void setString(const char *value);
-    char *getString();
+    void setString(std::string value);
+    std::string getString();
```
mdavidsaver: Be careful as there is potentially a string copy happening here.  Unless a returning a temporary, it is almost always preferable to return a const reference.

```
const std::string& getString();
```

Also note that changing return type between value and const reference generally doesn't require changes to API user code.

mdavidsaver:
Also, I notice that paramVal.h is installed, but as far as I can tell, not usable.  Is this intended to be a public API?

mdavidsaver:
As a note for the future.  I see two alternatives to a C union.  They would both require some thought about what should happen when, for example, setString is called on a Double paramVal.

1. Make paramVal an abstract base class, and it's set/get methods as virtual calls.  The appropriate derived class is selected in createParam.  Ine sub-class might keep the C union and handle all the scalar cases, and others for each non-POD and array case.

2. Manage storage w/ eg. std::vector<char> and cast it's pointer to the necessary type for each access.  This includes using placement new and manual destruction for non-POD types, and giving some thought to storage alignment.

Option 1 is more straightforward to understand, but will result in more source lines, and another level of indirection.

Option 2 is complicated to understand and thus has the potential for bugs.  (for the record, placement new is already being used in Base, so compatibility need not be a problem)

I'm be inclined towards option 1 unless there was some important performance regression.  Given how much indirection is already involved in, I doubt that this would come up.

@MarkRivers it might be worthwhile to glance at the 3 categories of generic value container listed at http://www.boost.org/doc/libs/1_62_0/doc/html/any.html

Having run across all three in the past, I would agree with this author that Discriminated types are the easiest to work with.

asyn/asynPortDriver/paramVal.h
MarkRivers: Done, thanks.  Also added const on setter.

asyn/asynPortDriver/paramVal.h
MarkRivers: Done.  I moved sval to be a non-pointer member outside the union.

MarkRivers:
> Also, I notice that paramVal.h is installed, but as far as I can tell, not usable. Is this intended to be a public API?

Thanks for pointing that out.  I have changed the Makefile so that paramErrors.h is now installed and paramVal.h is not.  paramVal is not intended to be a public API.  This change let me clean up the unittest, since it was redefining the contents of paramErrors.h.

ulrikpedersen:
I'll try to get a VxWorks 5.5 system setup to run the asynPortDriverTest... I do like the move to using STL containers but we're stuck with many VxWorks 5.5 systems around.  We've had issues with other STL code in the past so I just need to make sure asyn doesn't crash now.

ulrikpedersen:
I know I'm late to the party; but just to let you know that I've run both the testAsynPortDriver IOC and the asynPortDriverTest unittest on VxWorks 5.5 (and on RHEL6). 

No crashes and it looks like it works as expected.

Trace from asynPortDriverTest on VxWorks 5.5:

```
ioc258 -> asynPortDriverTest
1..53
# Basic parameter creation
ok  1 - portA->findParam(0, "int32", &idx1)==asynParamNotFound
ok  2 - portA->getParamName(0, 0, &name)==asynParamBadIndex
epicsTimeGetCurrent failed
asynPortDriver:getParamAlarmSeverity: port=portA error setting parameter 0 in list 0, bad index
ok  3 - portA->getParamAlarmSeverity(0, &sevr)==asynParamBadIndex
epicsTimeGetCurrent failed
asynPortDriver:getParamAlarmStatus: port=portA error setting parameter 0 in list 0, bad index
ok  4 - portA->getParamAlarmStatus(0, &sevr)==asynParamBadIndex
ok  5 - portA->createParam(0, "int32", asynParamInt32, &idx1)==asynSuccess
ok  6 - portA->findParam(0, "int32", &idx2)==asynSuccess
ok  7 - idx1==idx2
ok  8 - portA->getParamName(0, 0, &name)==asynSuccess
ok  9 - strcmp(name, "int32")==0
ok 10 - portA->getParamAlarmSeverity(0, &sevr)==asynSuccess
ok 11 - sevr==0
ok 12 - portA->getParamAlarmStatus(0, &sevr)==asynSuccess
ok 13 - sevr==0
epicsTimeGetCurrent failed
asynPortDriver:getDoubleParam: port=portA error getting parameter 0 int32, in list 0, wrong type
ok 14 - portA->getDoubleParam(0, idx1, &dval)==asynParamWrongType
ok 15 - portA->getIntegerParam(0, idx1, &ival)==asynParamUndefined
epicsTimeGetCurrent failed
asynPortDriver:createParam: port=portA error adding parameter int32 to list 0, parameter already exists.
ok 16 - portA->createParam(0, "int32", asynParamInt32, &idx2)==asynError
ok 17 - portA->createParam(0, "float64", asynParamFloat64, &idx1)==asynSuccess
ok 18 - portA->createParam(0, "uint32", asynParamUInt32Digital, &idx1)==asynSuccess
ok 19 - portA->createParam(0, "y", asynParamInt32, &idx1)==asynSuccess
ok 20 - portA->createParam(0, "z", asynParamFloat64, &idx1)==asynSuccess
ok 21 - portA->createParam(0, "more", asynParamFloat64, &idx1)==asynSuccess
ok 22 - portA->findParam(0, "more", &idx2)==asynSuccess
ok 23 - portA->createParam(0, "string", asynParamOctet, &idx1)==asynSuccess
ok 24 - portA->findParam(0, "int32", &idx1)==asynSuccess
epicsTimeGetCurrent failed
asynPortDriver:setDoubleParam: port=portA error setting parameter 0 in list 0, wrong type
ok 25 - portA->setDoubleParam(0, idx1, 4.2)==asynParamWrongType
ok 26 - portA->setIntegerParam(0, idx1, 42)==asynSuccess
ok 27 - portA->getIntegerParam(0, idx1, &ival)==asynSuccess
ok 28 - ival==42
ok 29 - portA->findParam(0, "string", &idx1)==asynSuccess
ok 30 - portA->setStringParam(0, idx1, "Testing 123")==asynSuccess
ok 31 - portA->getStringParam(0, idx1, sizeof(str)-1, str)==asynSuccess
ok 32 - strcmp(str, "Testing 123")==0
ok 33 - portA->getStringParam(0, idx1, stdString)==asynSuccess
ok 34 - stdString == "Testing 123"
ok 35 - portA->setStringParam(0, idx1, stdString)==asynSuccess
ok 36 - portA->getStringParam(0, idx1, stdString)==asynSuccess
ok 37 - stdString == "Hello world"
ok 38 - portA->findParam(0, "y", &idx1)==asynSuccess
ok 39 - client.registerInterruptUser(&int32cb)==asynSuccess
ok 40 - cbcount==0
ok 41 - lastint32==1234
ok 42 - client.read(&val)==asynParamUndefined
# int32cb() called with 55
ok 43 - client->read(&other)==asynSuccess
ok 44 - data==other
# int32cb() done
ok 45 - portA->callParamCallbacks()==asynSuccess
ok 46 - lastint32==55
ok 47 - cbcount==1
# int32cb() called with 43
ok 48 - client->read(&other)==asynSuccess
ok 49 - data==other
# int32cb() done
ok 50 - client.read(&val)==asynSuccess
ok 51 - val==43
ok 52 - lastint32==43
ok 53 - cbcount==2

    Results
    =======
       Tests: 53 
      Passed:  53 = 100.00%
value = 0 = 0x0
```

PR #41, opened 2017-01-27
Fix warning about deprecated auto_ptr

mark0n:


anjohnson:
-1

I don't believe this change is portable to all the OSs that Asyn currently runs on — even the Travis CI build fails because of it.

mark0n:
Sorry, I wasn't aware that you guys still support some targets that do not support C++11. Closing this for now.

PR #42, opened 2017-02-01
Add Optional LeCroy Mode to asyn TCP/IP Port Driver

shadowguy:
Hi, Mike Browne at SLAC added some code to drvAsynIPPort.c which handles serial communication with LeCroy oscilloscopes - Needed because these scopes use and expect a special header file on both sent and received data before each data block which asyn was unable to process before.  The code mostly ignores the header except to document the size of the block before the next header.

As this seems like a useful function and we have been using it at LCLS for almost a year I thought it would be a good addition for the community.  This is an optional mode that is off by default and must be explicitly turned on with an asynSetOption command.

Using the LeCroy mode is simple and just takes one line in the st.cmd file (or shell):
asynSetOption("portname", addr, "LeCroy", "Y")
replacing 'portname' and 'addr' with their values in your IOC.

To turn off LeCroy mode either comment out the line and restart or put in an "N":
asynSetOption("portname", addr, "LeCroy", "N")

Snip from the LeCroy manual explaining the header format:
lecroy

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPPort.c
@@ -94,6 +94,8 @@
 
 #define ISCOM_UNKNOWN (-1)
 
+#define LECROY /* For optional LeCroy mode */
+
 /*
  * This structure holds the hardware-specific information for a single
  * asyn link.  There is one for each IP socket.
--- a/asyn/drvAsynSerial/drvAsynIPPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPPort.c
@@ -94,8 +94,6 @@
 
 #define ISCOM_UNKNOWN (-1)
 
-#define LECROY /* For optional LeCroy mode */
-
 /*
  * This structure holds the hardware-specific information for a single
  * asyn link.  There is one for each IP socket.
@@ -124,14 +122,11 @@ typedef struct {
     asynInterface      common;
     asynInterface      option;
     asynInterface      octet;
-#ifdef LECROY
     int                lecroy;
     int                lecroy_error;
     int                lecroy_length;
-#endif /* LECROY */
 } ttyController_t;
 
-#ifdef LECROY
 typedef struct lecroy {
     unsigned char operation;
     unsigned char version;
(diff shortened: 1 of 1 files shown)


norumwe12:
Could this be handled in an interposed interface?  It seems pretty esoteric to place into the core driver.

MarkRivers:
I agree with Eric, this looks like a perfect case for using the asyn interposeInterface capability.

It is completely analogous with Modbus/TCP.  Modbus/TCP adds a header and CRC trailer to the standard Modbus message when it is transmitted over TCP.  I have handled this by using an pasynManager->interposeInterface to remove the header and CRC on read, and add them on write.  See https://github.com/epics-modules/modbus/blob/master/modbusApp/src/modbusInterpose.c for how I did it.  That code is more complex than you need, because it handles 3 different protocols: Modbus/TCP, Modbus/RTU and Modbus/ASCII.  Your interpose interface code would be considerably simpler.  The LeCroy specific code could then be in the same module as your LeCroy driver.

I really don't think this type of vendor-specific code belongs in the base drvAsynIPPort driver.

shadowguy:
Hi Eric, Mark, thanks for the replies.  TBH I was wondering the same thing but just thought I'd offer it.

I've already started going through the modbus code to adapt it, but there is one other significant difference.  Since this is a specific kind of device our IOC for this uses a streamdevice protocol file for the 'byte stream' driver.  Do you know if streamdevice (as written) would be able to use an asyn interposed interface?  Open to suggestions as well.

norumwe12:
StreamDevice works fine with interposed interfaces. For example, StreamDevice is often used with the asynInterposCOM interface which makes LAN/Serial adapters using the RFC 2117 protocol  appear like standard serial ports.

MarkRivers:
Yes, the interpose interface is transparent to the asyn client.  You do the following:

- Create the drvAsynIPPort "myPort"
- Create the interpose interface on "myPort" which interposes the asynOctet interface.
- Load database with StreamDevice support.  It connects to "myPort" on the asynOctet interface and does not know that you have an interposed the interface.

You actually probably do this all the time without even knowing it, because you are transparently using the asynEos interpose interface.  It handles the EOS processing if you enable it, which you normally do in the final 0 argument to drvAsynIPPortConfigure and drvAsynSerialPortConfigure.

norumwe12:
Forgot to mention in my previous message that the asynInterposeCOM code might be a better place to start from since its actions are closer (than the MODBUS interposed interface) to what the LeCroy layer will need.  That, and the fact that asynInterposeCOM is a whole lot smaller.

Issue #46, opened 2017-02-08
Reporting that the parameter value has changed to the driver


hinxx:
I have a proposal for an enhancement.

I'm working with hardware that needs to have parameters updated at certain point in time. For example if DSP on FPGA is in 'armed' mode and processing is taking place parameter updates, that involve setting a register value, need to be delayed to the time after the DSP processing has finished.

I was thinking that this might be an addition to the asynPortDriver, more specifically to the paramList class by introducing 'value changed' status. It would function in a similar manner as 'flags' member currently works. Upon PV hence param value change, 'value changed' status is set. Then asynPortDriver derived class can check if param 'value changed' status is set and can act accordingly; use the parameter value when needed (update the FPGA register when safe). Afterwards it would reset the 'value changed' status.

asynPortDriver derived class is in total control of the 'value changed' status, because it needs to knows and might even ignore it if it wants to..

Does this make sense?

hinxx:
Here is the initial stab at this - #47

hinxx:
Second attempt can be found here - #48.

MarkRivers:
Hinko found another solution to this problem was found and said he no longer needs this change.

PR #47, opened 2017-02-15
Introduce clearParamValueChanged() and hasParamValueChanged() methods.

hinxx:
Allows the user to check if the parameter value changed. Useful if
acting on hardware must be a delayed action.

I still need to do more testing in order to better verify the functionality.

See also #46.

### DIFF
--- a/asyn/asynPortDriver/asynPortDriver.cpp
+++ b/asyn/asynPortDriver/asynPortDriver.cpp
@@ -73,6 +73,9 @@ public:
     asynStatus getAlarmStatus(int index, int *alarmStatus);
     asynStatus setAlarmSeverity(int index, int alarmSeverity);
     asynStatus getAlarmSeverity(int index, int *alarmSeverity);
+    asynStatus clearValueChanged(int index);
+    asynStatus setValueChanged(int index);
+    asynStatus hasValueChanged(int index, int *changed);
     void report(FILE *fp, int details);
 
 private:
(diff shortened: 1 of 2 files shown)


anjohnson:
This looks like a rather inefficient way to store the valueChanged data, since all of the related methods have to iterate through the vector and do an unsigned comparison for every iteration. I don't know the asynPortDriver code internals at all, but this approach doesn't seem right to me.

Why not define valueChanged to be a vector<bool> and use the parameter number directly to index into that vector? This gives immediate access to the valueChanged status instead of having to search a list for the specific parameter index number. I accept that most of the time there probably won't be very many entries in your list, but I suspect that a simple index into a vector<bool> will always be faster than setting up the iteration loop, and will be deterministic (always taking the same time to return a result).

Some drivers might also want a way to clear (or set) all the valueChanged bits at once, after FPGA initialization or to trigger writing all the parameters to the FPGA after a chip reset perhaps.

hinxx:
I've used the paramList::setFlag() as an example for providing this functionality. I figured that if that method is acceptable one could do something similar.

Nevertheless, I'm willing to optimize as you pointed out, if the whole value changed functionality is reasonable.

keenanlang:
You are automatically adding entries into valueChanged every time registerParameterChange is called by the driver, however I don't see any removal of entries without requiring the programmer to use the manual calls provided. That means that every single existing asynPortDriver that doesn't use the valueChanged functionality is going to start leaking memory every time a value is changed.

hinxx:
See #48 for improved version.

PR #48, opened 2017-02-18
Introduce new value checks.

hinxx:
Second stab at providing clients with capability of checking if parameter value has changed.

This attempt tries to address performance issues and possible memory leaks compared to #47.

Functionality moved to paramVal class by introducing isValueNew() and resetValueNew() methods.
asynPortDriver methods clearParamValueNew() and isParamValueNew() are added that call above mentioned paramVal methods.
User derived asynPortDriver class is responsible for calling the clearParamValueNew().
Existing asynPortDriver derived classes should not be affect by this change, in other words clearParamValueNew() and isParamValueNew() can be ignored.

Changes are under test with a concrete use case as we speak.

See also #46

### DIFF
--- a/asyn/asynPortDriver/asynPortDriver.cpp
+++ b/asyn/asynPortDriver/asynPortDriver.cpp
@@ -73,6 +73,9 @@ public:
     asynStatus getAlarmStatus(int index, int *alarmStatus);
     asynStatus setAlarmSeverity(int index, int alarmSeverity);
     asynStatus getAlarmSeverity(int index, int *alarmSeverity);
+    asynStatus getValueNew(int index, bool *status);
+    asynStatus resetValueNew(int index);
+
     void report(FILE *fp, int details);
 
 private:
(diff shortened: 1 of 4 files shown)


MarkRivers:
Can you explain the use case?  Would it be possible to add methods to check and set the existing Changed flag rather than adding the New flag.  How do Changed and New differ?

hinxx:
Thanks for looking into this one @MarkRivers. I have to remind myself of the exact intent since it was so long ago.. here goes.

The use case would be that when the camera is exposing / acquiring image one does not want to change camera settings - Andor SDK would complain to the user that acquisition needs to be stopped first. In this case ADAndor driver ignores users new setting AFAIK (maybe that changed?).
This pull request was my attempt to allow user to set new parameter and not worry about the current state of the detector.

User would change the value of the PV, then writeInt32() simply calls setIntegerParam() and returns.
After the image acquisition on data thread finished, check is made if param value has changed, if so write to hardware. This should be safe to do since we are between the "exposures". Params that are not sent to hardware would be already updated during exposure - but might collide with possible cases where such params need to remain unchanged after the image is acquired (if doing some processing or stats?).

I was seeing this useful especially for ADC drivers based AD where FPGA registers should not be changed while DSP is active in firmware. Very common in pulsed machines, what I'm currently dealing with.

Whether existing Changed flag can be used for this use case is something I would need to look more deeply into. Can you tell from the use case above if this is doable, since you know internal way better than I do?

hinxx:
@Markrivers, do you think this pull request can make it into mainstream? If not, I will consider doing something similar locally, and remove the pull request.

MarkRivers:
Sorry for the delay on this.  I've looked at it again, and this seems like there are cases where it could be useful.  Are you still interested in having it merged?

hinxx:
After talking to out FPGA guys we decided that FPGA firmware should have registers shadowed and allow software to trigger the update of actual registers at a known point in time. This way software does not have to cache new user values.

That being said, I think you do not have to merge this PR.

MarkRivers:
Closing without merging per message from @hinxx above.

PR #51, opened 2017-03-02
in ipSNCServer.st, renamed pVar parameter to avoid name collisions

bfrk:
I simply changed it to all lowercase. This change is fully forward and
backward compatible with all supported sequencer versions. Yet it allows me
to significantly simplify the code generator in version 2.3.

### DIFF
--- a/testIPServerApp/src/ipSNCServer.st
+++ b/testIPServerApp/src/ipSNCServer.st
@@ -24,9 +24,9 @@ char *registrarPvt; /* This is really void* */
 char *eventId;      /* Thus is really epicsEventId */
 
 /* Define escaped C functions at end of file */
-%% static void initialize(SS_ID ssId, struct UserVar *pVar);
-%% static int readSocket(SS_ID ssId, struct UserVar *pVar);
-%% static int writeSocket(SS_ID ssId, struct UserVar *pVar);
+%% static void initialize(SS_ID ssId, struct UserVar *pvar);
+%% static int readSocket(SS_ID ssId, struct UserVar *pvar);
+%% static int writeSocket(SS_ID ssId, struct UserVar *pvar);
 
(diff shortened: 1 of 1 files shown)


MarkRivers:
There are 10 other .st files in synApps with 971 lines containing the pVar variable.  Do these all need to be changed?  Why is there now a name conflict with pVar?

bfrk:
> Why is there now a name conflict with pVar?

tl;dr: the conflict is in there since seq version 2.2 but I used an ugly hack to work around it. I thought I could get rid of this hack in version 2.3 but I see now that the cost to convert user code might be higher than I expected. So I think I will retain the hack.

Somewhat longer version: From 2.1 to 2.2 there was one important change: the functions that snc generates get only one extra parameter, not two. In 2.1 those were `SS_ID ssId, UserVar pVar`, whereas from 2.2 onward only the `SS_ID ssId` remains. I also renamed the name of the parameter to `seqg_env`. I made this change because it is easy to recover the `pVar` from the `ssId`, since the `ssId`contains this pointer anyway; it just needs to be casted to the generated struct. For this I added a macro `seqg_var` that does the ugly pointer casting and an alias for it named `pVar`for compatibility. However, while this macro works fine for accessing variables, it gets in the way when you /declare/ something with the name `pVar`, which is what your program does in the parameter list for the C functions. To make seq-2.2 compatible with such uses of the identifyer `pVar` I had to `#define` and `#undef` the macro in the generated code, depending on whether escaped C code is being output (where pVar might be declared and then used as a function parameter) or whether it is actual SNL code that gets converted to C. I had more or less forgotten about this whole story and wondered how to get rid of this extra complication which led me to propose this pull. But, as I said above, there is no real need for it, so I hereby withdraw the pull request.

PR #54, opened 2017-03-16
Added getParamType methods

jgnrendra3:
As near as I can tell, there is no (reliable) way to get the asynParamType for a parameter without this function.  Adding this will fill that hole without forcing clients to know things thath might change regarding the internal elements of multiple classes (e.g. the structure of the paramList class, which is not in any header file).


jgnrendra3:
Sorry about that.  The editor I normally use is configured to eliminate 
trailing spaces.

I resubmitted it with the existing trailing spaces intact.

Mark Davis
NSCL/FRIB Control Systems Software Engineer
davism50@msu.edu

PR #59, opened 2017-08-08
fix compile errors on systems where epicsInt32 ==long int

sbaily:
Warnings have not been fixed.  On my system long int and int are both 32 bits, but epicsInt32 is mapped to long int instead of int, this causes build errors whenever pointers are referred to with different data types. My changes correct these errors.

### DIFF
--- a/asyn/asynPortDriver/asynPortDriver.cpp
+++ b/asyn/asynPortDriver/asynPortDriver.cpp
@@ -57,7 +57,7 @@ public:
     asynStatus setDouble(int index, double value);
     asynStatus setString(int index, const char *string);
     asynStatus setString(int index, const std::string& string);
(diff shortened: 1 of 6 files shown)


MarkRivers:
I'm curious what architectures you are using that have this problem?

sbaily:
This is a NIOS-II soft-core processor on an Altera FPGA, and Jeff Hill wrote the BSP for RTEMS.  On this system int and long int are both 32 bits, but epicsInt32 maps to long int rather than int.  I suppose this could be changed instead of changing asyn.  However, I noticed that asynPortDriver.h used epicsInt32* as the type for the same parameter that was implementing using the int* in asynPortDriver.cpp, and figured it should at least be consistent.

Issue #56, opened 2017-05-01
Race condition with info tag asyn:READBACK


MarkRivers:
@mdavidsaver wrote:
> Something doesn't look right in processAo() [1].  To correctly do a
bi-directional (setting and readback) record your process function has
to know whether it's being called with a new setting, or to pull an
updated readback.  The fact that record processing (by design) lacks
this contextual information is why the standard device supports can't do
this.
Simply checking the internal readback FIFO allows a race which can
ignore a new value from CA.  I'm fairly confident that this is what
Garth is seeing.
I don't immediately see where to fix this in the context of asyn.

and
> In interruptCallbackOutput() [1], instead of scanOnce() use a custom
CALLBACK like:
```
static void ProcessCallback(CALLBACK *pcallback)
 {
     dbCommon *pRec;
     devPvt *pPvt;
 
     callbackGetUser(pRec, pcallback);
     if (!pRec) return;
     pPvt = (devPvt *)pRec->dpvt;
     dbScanLock(pRec);
     pPvt->updateReadback = 1;
     dbProcess(pRec);
     pPvt->updateReadback = 0;
     dbScanUnlock(pRec);
 }
```
> Where 'updateReadback' is a new flag which can be tested in processAo()
to determine why processing has been started: a new setting value
(updateReadback=0) or a new readback (updateReadback=1).

MarkRivers:
This issue has been fixed.

PR #60, opened 2017-10-31
Fix processing of output records with asyn:READBACK tag on driver callbacks

MarkRivers:
This attempts to fix the race condition identified in #56.  Rather than just using the readback value if the ring buffer has any entries it uses callbackRequest to signal to the process routine that this is a driver callback.  This allows callback processing and normal record processing to be interleaved.

I do have a question about how I implemented this.  In the code @mdavidsaver proposed in #56 he called dbProcess in the callback function.  I am calling the prset->process().  Which is correct?  Can someone explain what happens if the reacback callback occurs while the record is currently in the middle of normal asynchronous processing, i.e. PACT is true waiting for asynchronous completion?

### DIFF
--- a/asyn/devEpics/devAsynFloat64.c
+++ b/asyn/devEpics/devAsynFloat64.c
@@ -91,6 +91,7 @@ static long createRingBuffer(dbCommon *pr);
 static void processCallbackInput(asynUser *pasynUser);
 static void processCallbackOutput(asynUser *pasynUser);
 static void outputCallbackCallback(CALLBACK *pcb);
+static int  getCallbackValue(devPvt *pPvt);
 static void interruptCallbackInput(void *drvPvt, asynUser *pasynUser,
                 epicsFloat64 value);
 static void interruptCallbackOutput(void *drvPvt, asynUser *pasynUser,
@@ -418,11 +419,14 @@ static void outputCallbackCallback(CALLBACK *pcb)
     callbackGetUser(pPvt, pcb);
     {
         dbCommon *pr = pPvt->pr;
-        struct rset *prset = (struct rset *)pr->rset;
         dbScanLock(pr);
         pPvt->newOutputCallbackValue = 1;
-        (prset->process)(pr);
-        pPvt->newOutputCallbackValue = 0;
+        dbProcess(pr);
+        if (pPvt->newOutputCallbackValue != 0) {
+            /* We called dbProcess but the record did not process, perhaps because PACT was 1 
+             * Need to remove ring buffer element */
+            getCallbackValue(pPvt);
+        }
         dbScanUnlock(pr);
     }
 }
@@ -463,6 +467,7 @@ static int getCallbackValue(devPvt *pPvt)
                                             pPvt->pr->name,pPvt->result.value);
         ret = 1;
     }
+    pPvt->newOutputCallbackValue = 0;
     epicsMutexUnlock(pPvt->ringBufferLock);
     return ret;
 }
--- a/asyn/devEpics/devAsynFloat64.c
+++ b/asyn/devEpics/devAsynFloat64.c
@@ -414,16 +414,17 @@ static void interruptCallbackOutput(void *drvPvt, asynUser *pasynUser,
 
 static void outputCallbackCallback(CALLBACK *pcb)
 {
-    devPvt *pPvt;
+    devPvt *pPvt; 
     callbackGetUser(pPvt, pcb);
-    dbCommon *pr = pPvt->pr;
-    struct rset *prset = (struct rset *)pr->rset;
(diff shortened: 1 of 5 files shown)


mdavidsaver:
A short answer is that ```outputCallbackCallback()``` needs to use ```dbProcess()``` because ```scanOnce()``` does.

A longer answer is that ```dbProcess()``` should be used to __start__ record processing.  Directly calling ```rset->process()``` is done to __complete__ asynchronous record processing.  In the case of asyn interrupts, processing is being started, so ```dbProcess()``` is appropriate.

To directly answer your question.  Calling ```dbProcess()``` on an active record is a no-op as this is where PACT==1 is tested.

https://github.com/epics-base/epics-base/blob/3.16/src/ioc/db/dbAccess.c#L522

This is also were the scan disable test is done (aka. SDIS/DISV/DISA).

MarkRivers:
> To directly answer your question. Calling dbProcess() on an active record is a no-op as this is where PACT==1 is tested.

I think this means my code needs to be changed.  This is because by the time dbProcess is called the driver has already pushed a value onto the ring buffer, but since the record won't process if PACT=1 then that value won't be properly removed from the ring buffer.  It appears that dbProcess does not return an error flag if PACT is 1.  So I think I need to test if PACT=1 in my callback, and if it is then remove the value from the ring buffer and not call dbProcess?  Or should I call dbProcess to let it possbily put the record into SCAN alarm?

mdavidsaver:
You're right, the FIFO complicates this.  And no, dbProcess() doesn't tell you if the scan was disabled.  My first thought is to have the device support clear newOutputCallbackValue and check this in the callback.  Then pop from the FIFO if the flag wasn't cleared.  This seems like it will led to some code duplication though.

MarkRivers:
> My first thought is to have the device support clear newOutputCallbackValue and check this in the callback. Then pop from the FIFO if the flag wasn't cleared. This seems like it will led to some code duplication though.

I've implemented this.  newOutputCallbackValue is now cleared in getCallbackValue().  If that was not called because dbProcess did not process the record then getCallbackValue() is called in the callback, which both pops the FIFO and clears the flag.  I made the same fix in devBusyAsyn in the busy module.

MarkRivers:
I have found that there is still a problem.  Output readbacks should change for all combinations of the following settings.
1) Driver is synchronous (ASYN_CANBLOCK not set)
2) Driver is asynchronous (ASYN_CANBLOCK is set)
3) Callbacks are done from a background thread in the driver
4) Callbacks are done immediately during the write operation to the driver from device support.  For example the record writes the value 1.5 to the driver, but this is not an allowed value so the driver does a callback with the valid value 1 during the write operation.
5) The driver does a single callback
6) The driver does multiple callbacks quickly, either in setting 3) or 4) above

I have written a new test application in asyn, testOutputCallbackApp which allows testing all combinations of these 6 settings above.  This application shows that combining settings 2 and 4 does not work, i.e. an asynchronous driver with callbacks being done immediately in the call to the write function in the driver.  This is a serious problem, particularly because the same logic for the bo record in this code, and the busy record must process again with the value 0 if the value 1 was written by the driver and the driver got done immediately.

The reason for the problem is that the callbackRequest to process the record again is being done in the interrupt callback function in device support.  But if the record is synchronous PACT is 1 at this point.  If the callback runs before the record completes the asynchronous processing that caused the write operation, then dbProcess will not process the record again.  I have added an asynPrint with ASYN_TRACE_WARNING to the device support when this happens, which indeed it does.

It seems that we need to delay calling callbackRequest until the record is done processing the second phase if the driver callback happened when PACT=1.  

@mdavidsaver and @anjohnson how would you suggest to arrange this?

On a perhaps related note, @mdavidsaver said that 
> Directly calling rset->process() is done to complete asynchronous record processing. 

But in the asyn device support it actually does not call rset->process, it calls callbackRequestProcessCallback.  I suspect that Marty did this so that the record processing would not be done in the asyn port driver thread, but rather in one of the generic callback tasks.  This should be OK, correct?

MarkRivers:
I have come up with a solution to the above problems.  
- In the interrupt callback for output records test if PACT=1.  
- If so this is an asynchronous record which is in the middle of processing.  
    - Don't call callbackRequest
    - Instead increment a new numDeferredOutputCallbacks counter
- At the end of the process() function in device support test if numDeferredOutputCallbacks is > 0
- If so, call callbackRequest to process the record again and decrement the counter.

I also found that in the processCallbackInput and processCallbackOutput functions that dbScanLock was not being called.  I think it should be since these functions are called by asynManager to do the actual I/O, so the scan lock won't already be held.  They are testing PACT and setting values in the device private structure, so the lock is needed.

PR #64, opened 2017-11-27
issue messages during auto-connect only on status change

bfrk:
This changes asynManager and the vxi-11 port driver.


bfrk:
Will rebase before re-sending pull requerst. I accidentally used an old fork.

PR #66, opened 2017-12-07
devEpics: add ASLO/AOFF/SMOO conversion to ai/ao float64

ralphlange:
This is adding ASLO/AOFF conversion (and SMOO handling for ai) to the float64 support for ai/ao records.

Record support code always converts between RVAL (integer) and VAL (double), so it can't be used in cases like this, where the support directly reads/writes the VAL field. This PR adds support for the minimal set of ASLO/AOFF/SMOO conversions. It does not go all the way adding ESLO/EOFF linear and breakpoint table conversions, avoiding excessive code bloat.

### DIFF
--- a/asyn/devEpics/devAsynFloat64.c
+++ b/asyn/devEpics/devAsynFloat64.c
@@ -581,7 +581,7 @@ static long initAo(aoRecord *pao)
         /* ASLO/AOFF conversion */
         if (pao->aslo != 0.0) val64 *= pao->aslo;
         val64 += pao->aoff;
-        pao->val = value;
+        pao->val = val64;
         pao->udf = 0;
     }
     pasynFloat64SyncIO->disconnect(pPvt->pasynUserSync);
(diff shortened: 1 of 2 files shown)


ralphlange:
I wanted to add a test to testEpics, but I had to realize I don't understand your test setup and procedure good enough to do that.

MarkRivers:
I think you may be missing a conversion at the top of processAo().  Doesn't it need to do the conversion on the readback value which happens for records with the asyn:READBACK field set?

ralphlange:
Ehm... yes.

ralphlange:
Fixed/added. Thanks for the review!

MarkRivers:
This did not compile on WIN32 or vxWorks because finite() requires epicsMath,h on those architectures.  I added it and now it works.  APS Jenkins comes in very handy!

ralphlange:
Duh. Thank you!

PR #69, opened 2018-02-01
XON/XOFF flow control setting available in IOC shell

daykin:
We needed this functionality for use with a Rubidium frequency standard. It works if specified in the IOC shell, but I have not yet implemented it as a setting in the asyn record.

### DIFF
--- a/asyn/miscellaneous/asynInterposeCom.c
+++ b/asyn/miscellaneous/asynInterposeCom.c
@@ -52,7 +52,7 @@
 #define CPO_SET_STOPSIZE         4   /* Comand port option set stop size */
 #define CPO_SET_CONTROL          5   /* Comand port option set control mode */
 # define CPO_CONTROL_NOFLOW        1   /* No flow control */
-# define CPO_CONTROL_XONOFF       2   /* XON/XOFF Flow control*/
+# define CPO_CONTROL_IXON       2   /* XON/XOFF Flow control*/
 # define CPO_CONTROL_HWFLOW        3   /* Hardware flow control */
 #define CPO_SET_LINESTATE_MASK  10 /* Comand port option set linestate mask */
 #define CPO_SET_MODEMSTATE_MASK 11 /* Comand port option set modemstate mask */
@@ -565,7 +565,7 @@ setOption(void *ppvt, asynUser *pasynUser, const char *key, const char *val)
     }
     else if (epicsStrCaseCmp(key, "crtscts") == 0) {
         xBuf[0] = CPO_SET_CONTROL;
-        if (pinterposePvt->flow == CPO_CONTROL_XONOFF){
+        if (pinterposePvt->flow == CPO_CONTROL_IXON){
             epicsSnprintf(pasynUser->errorMessage, pasynUser->errorMessageSize,
                             "XON/XOFF already set. Now using RTS/CTS.");
         }
@@ -579,14 +579,14 @@ setOption(void *ppvt, asynUser *pasynUser, const char *key, const char *val)
         status = sbComPortOption(pinterposePvt, pasynUser, xBuf, 2, rBuf);
         if (status == asynSuccess) pinterposePvt->flow = rBuf[0] & 0xFF;
     }
-    else if (epicsStrCaseCmp(key, "xonoff") == 0){
+    else if (epicsStrCaseCmp(key, "ixon") == 0){
         xBuf[0] = CPO_SET_CONTROL;
         if (pinterposePvt->flow == CPO_CONTROL_HWFLOW){
             epicsSnprintf(pasynUser->errorMessage, pasynUser->errorMessageSize,
                              "RTS/CTS already set. Now using XON/XOFF.");
         }
         if      (epicsStrCaseCmp(val, "n") == 0) xBuf[1] = pinterposePvt->flow;
-        else if (epicsStrCaseCmp(val, "y") == 0) xBuf[1] = CPO_CONTROL_XONOFF;
+        else if (epicsStrCaseCmp(val, "y") == 0) xBuf[1] = CPO_CONTROL_IXON;
         else {
             epicsSnprintf(pasynUser->errorMessage, pasynUser->errorMessageSize,
                                                                   "Bad option value");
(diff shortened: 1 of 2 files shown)


mark0n:
Note: This adds XON/XOFF support for the [RFC2217](https://tools.ietf.org/html/rfc2217) mode in which the IOC remotely configures the serial-port settings of a device server. Tested with a [MOXA NPort 6650](https://www.moxa.com/product/NPort_6650.htm).

MarkRivers:
Sorry for delay, I looked at this the other day and meant to ask one question.  The key is called "cxonxoff".  Why the "c", would "xonxoff" be better?

mark0n:
I guess @daykin added the "c" since the [parameter for RTS/CTS is called "crtscts"](https://github.com/epics-modules/asyn/blob/master/asyn/miscellaneous/asynInterposeCom.c#L565). Not sure why that one has the "c". Maybe a typo?

norumwe12:
The RTS/CTS name is from POSIX termios:
>     Values of the c_cflag field describe the basic terminal hardware control,
>      and are composed of the following masks.  Not all values specified are
>      supported by all hardware.
> 
>            CSIZE       /* character size mask */
>            CS5         /* 5 bits (pseudo) */
>            CS6         /* 6 bits */
>            CS7         /* 7 bits */
>            CS8         /* 8 bits */
>            CSTOPB      /* send 2 stop bits */
>            CREAD       /* enable receiver */
>            PARENB      /* parity enable */
>            PARODD      /* odd parity, else even */
>            HUPCL       /* hang up on last close */
>            CLOCAL      /* ignore modem status lines */
>            CCTS_OFLOW  /* CTS flow control of output */
>            CRTSCTS     /* same as CCTS_OFLOW */
>            CRTS_IFLOW  /* RTS flow control of input */
>            MDMBUF      /* flow control output via Carrier */

MarkRivers:
Since termios.h does not seem to have cxonxoff let's just use xonoff.

anjohnson:
My linux manpage for termios lists the following as constants for use with `c_iflag`:
```
       IXON   Enable XON/XOFF flow control on output.

       IXANY  (XSI) Typing any character will restart  stopped  output.   (The
              default is to allow just the START character to restart output.)

       IXOFF  Enable XON/XOFF flow control on input.
```
Note that it has independent controls for the input and output directions. The `IXANY` control is useful if the stream is talking to a human on a terminal though so may not be appropriate for Asyn.

norumwe12:
Right.
When this capability is added to the ASYN RFC2217 documentation I’d like to request the caveat that XON/XOFF flow control is less robust than RTS/CTS and that the latter be used if at all possible.  I’ve had way too much experience with device communication locking up because an XON got lost somewhere.

daykin:
I will change the keyword to "xonoff".

>I’d like to request the caveat that XON/XOFF flow control is less robust than RTS/CTS and that the latter be used if at all possible. 

If anyone ends up needing (in)outbound-only flow control, etc, we can discuss/add it later.

MarkRivers:
I just looked at the drvAsynSerialPirt driver. The options there are called ixon and ixoff. We should use the same option names in this driver. I am not sure if this is ixon or ixoff? Please also update asynDriver.html and RELEASE_NOTES.html.

daykin:
I was not sure what just 'ixon' or 'ixoff' imply as they relate to this standard. RFC2217 calls for specifying inbound only, or inbound and outbound. all of them both use XON and XOFF control signals.

MarkRivers:
The termios standard says this for serial ports:
```
       IXON   Enable XON/XOFF flow control on output.

       IXANY  (XSI) Typing any character will restart stopped output.  (The
              default is to allow just the START character to restart out‐
              put.)

       IXOFF  Enable XON/XOFF flow control on input.
```
RFC2217 says this:
```
IAC SB COM-PORT-OPTION SET-CONTROL <value> IAC SE
       This command is sent by the client to the access server to set
       special com port options. The command can also be sent to query
       the current option value. The value is one octet (byte). The
       value is an index into the following value table:

           Value      Control Commands
             0           Request Com Port Flow Control Setting
                           (outbound/both)
             1           Use No Flow Control (outbound/both)
             2           Use XON/XOFF Flow Control (outbound/both)
             3           Use HARDWARE Flow Control (outbound/both)
             4           Request BREAK State
             5           Set BREAK State ON
             6           Set BREAK State OFF
             7           Request DTR Signal State
             8           Set DTR Signal State ON
             9           Set DTR Signal State OFF
            10           Request RTS Signal State
            11           Set RTS Signal State ON
            12           Set RTS Signal State OFF
            13           Request Com Port Flow Control Setting (inbound)
            14           Use No Flow Control (inbound)
            15           Use XON/XOFF Flow Control (inbound)
            16           Use HARDWARE Flow Control (inbound)
            17           Use DCD Flow Control (outbound/both)
            18           Use DTR Flow Control (inbound)
            19           Use DSR Flow Control (outbound/both)
            20-127       Available for Future Use

       Discussion: Flow control options were divided into inbound and
                   outbound to take full advantage of existing
                   programming interfaces and access server
                   capabilities.

       Discussion: The outbound values should set flow control for both
                   outbound and inbound.  If inbound is to be, or can
                   be, set separately it should be done after the
                   setting of the outbound value.

       Discussion: If the access server is not able to set inbound flow
                   control differently from the outbound flow control,
                   it should ignore the inbound flow control commands
                   and set the flow control option based on the outbound
                   flow control commands only.
```
In the case of a serial port it is clear what outbound and inbound mean.  In the case of a terminal server it is not clear to me.  Does outbound refer to the direction from the terminal server to the device, or from the terminal server to the computer?  In any event I think we should use the strings ixon and ixoff using the mapping which makes sense based on the termios direction definitions.  This should be explained in asynDriver.html.

norumwe12:
I’m pretty sure that the direction is from the terminal server to the device — as if the serial port were directly connected to the computer — since the handshaking characters appear on the serial line and not the network link.

daykin:
With this in mind, I will change to IXON, since RFC2217 specifies the 'outbound' port option implies both inbound and outbound flow control. Does this sound good?

MarkRivers:
Yes, that sounds good, thanks.

daykin:
Documentation and naming is updated, is there anything else that needs to be done?

Issue #70, opened 2018-02-07
Uninitialized value in paramVal


mark0n:
When parameters get set for the first time like in
```
paramVal parVal("myParam", asynParamUInt32Digital);
parVal.setUInt32(42, 0, 0);
```
this results in the following issues being reported by Valgrind:
```
==7929== Conditional jump or move depends on uninitialised value(s)
==7929==    at 0x10AE88: paramVal::setUInt32(unsigned int, unsigned int, unsigned int) (paramVal.cpp:171)
==7929==    by 0x10AF8E: (anonymous namespace)::testA() (asynPortDriverTest.cpp:57)
==7929==    by 0x10A7E9: main (asynPortDriverTest.cpp:171)
==7929== 
==7929== Invalid read of size 8
==7929==    at 0x5F4B2A7: create_thread (createthread.c:116)
==7929==    by 0x5F4CFC7: pthread_create@@GLIBC_2.2.5 (pthread_create.c:822)
==7929==    by 0x5112F2E: epicsThreadCreate (osdThread.c:516)
==7929==    by 0x4E79F66: asynPortDriver::initialize(char const*, int, int, int, int, int, int, int) (asynPortDriver.cpp:3384)
==7929==    by 0x4E7ABCD: asynPortDriver::asynPortDriver(char const*, int, int, int, int, int, int, int) (asynPortDriver.cpp:3249)
==7929==    by 0x10AFCE: (anonymous namespace)::testA() (asynPortDriverTest.cpp:61)
==7929==    by 0x10A7E9: main (asynPortDriverTest.cpp:171)
==7929==  Address 0x7200028 is 72 bytes inside a block of size 166 free'd
==7929==    at 0x4C30D18: free (vg_replace_malloc.c:530)
==7929==    by 0x51125D8: start_routine (osdThread.c:406)
==7929==    by 0x5F4C7FB: start_thread (pthread_create.c:465)
==7929==    by 0x59F6B5E: clone (clone.S:95)
==7929==  Block was alloc'd at
==7929==    at 0x4C31A1E: calloc (vg_replace_malloc.c:711)
==7929==    by 0x5111EF2: create_threadInfo (osdThread.c:161)
==7929==    by 0x5111EF2: init_threadInfo (osdThread.c:180)
==7929==    by 0x5112EE1: epicsThreadCreate (osdThread.c:511)
[...]

PR #71, opened 2018-02-07
Fix uninitialized value in paramVal

mark0n:
This fixes issue #70.


MarkRivers:
Is there still a problem for parameters of type asynParamOctet, i.e. sval field or is that OK?

mark0n:
That should be OK since it of type `std::string`. The the `string::string` constructor will initialize it to `""`.

PR #73, opened 2018-05-17
Fix typo in devAsynInt32 interruptCallbackOutput's functionName

ebernalru:


mark0n:
We could avoid issues like this if we would populate these variables automatically. Unfortunately the corresponding macros (`__func__` etc.) are compiler specific. However, Boost provides [a solution that might work for us](https://www.boost.org/doc/libs/1_67_0/boost/current_function.hpp). @MarkRivers does this cover all compilers Asyn officially supports?

anjohnson:
In Base-3.16.1 we added the following to the EPICS compilerDependencies.h header file:
```
#ifndef EPICS_FUNCTION
#if (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)) || (defined(__cplusplus) && __cplusplus>=201103L)
#  define EPICS_FUNCTION __func__
#else
/* Expands to a 'const char*' which describes the name of the current function scope */
#  define EPICS_FUNCTION ("<unknown function>")
#endif
#endif
```
There are other compiler-specific headers included first that may also define `EPICS_FUNCTION` so for 3.16 and later I would recommend using that.

mark0n:
@anjohnson sounds reasonable. However, in cases like this @MarkRivers usually points out that Asyn can be used without EPICS ;-)

anjohnson:
@mark0n Actually he usually says it can be used without an EPICS IOC; it still relies on EPICS for the build system and uses libCom to provide OS-independence.

MarkRivers:
The problem I see is that neither the Boost version nor the EPICS 3.16 version is going to work on Visual Studio.  In Boost the tests that might work are these:
```
#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)

# define BOOST_CURRENT_FUNCTION __func__

#elif defined(__cplusplus) && (__cplusplus >= 201103)

# define BOOST_CURRENT_FUNCTION __func__

#else

```

EPICS 3.16 uses this test which is basically the same:
```
#if (defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)) || (defined(__cplusplus) && __cplusplus>=201103L)
#  define EPICS_FUNCTION __func__

```
@anjohnson  did you test to see if this works on Visual Studio?

From my reading of the MSDN docs I could find online no version of Visual Studio defines `__STDC_VERSION__` so that test won't work.  

`__cplusplus` is also not going to work because this is what MSDN says about it:

> The MSVC compiler’s definition of the __cplusplus predefined macro leaps ahead 20 years in Visual Studio 2017 version 15.7 Preview 3. This macro has stubbornly remained at the value “199711L”, indicating (erroneously!) that the compiler conformed to the C++98 Standard. 

...

> The version reported by the __cplusplus macro also depends upon the standard version switch used. If you’re compiling in C++14 mode the macro will be set to “201402L”. If you compile in C++17 mode the macro will be set to “201703L”. And the /std:c++latest switch, used to enable features from the Standard currently in development, sets a value that is more than the current Standard. This chart shows the values of the __cplusplus macro with different switch combinations:

/Zc:__cplusplus switch | /std:c++ switch | __cplusplus value
-- | -- | --
Zc:__cplusplus | Currently defaults to C++14 | 201402L
Zc:__cplusplus | /std:c++14 | 201402L
Zc:__cplusplus | /std:c++17 | 201703L
Zc:__cplusplus | /std:c++latest | 201704L
Zc:__cplusplus- (disabled) | Any value | 199711L
Zc:__cplusplus not specified | Any value | 199711L

So all versions of Visual Studio before Studio 2017 version 15.7 Preview 3 `__cplusplus` is 199711L, and the  `__cplusplus` tests in Boost and EPICS 3.16 will also fail.

The good news is that Visual Studio 2013 and higher do define `__func__`  so it can be used.  But I think the tests are going to need to be modified to test specifically for Visual Studio and which version.

mark0n:
How about this?
```
#elif defined(__STDC_VERSION__) && (__STDC_VERSION__ >= 199901)
# define BOOST_CURRENT_FUNCTION __func__
#elif defined(__cplusplus) && (__cplusplus >= 201103)
# define BOOST_CURRENT_FUNCTION __func__
#elif defined(_MSCVER)
# define BOOST_CURRENT_FUNCTION __FUNCTION__
#else
```
I don't have MSVC available, @MarkRivers could you please try this?

MarkRivers:
I'm on travel, I will test it early next week.

anjohnson:
In addition to the excerpt I posted, 3.16 also has compiler-specific headers that get included first, and I’m pretty sure that file will already take care of defining EPICS_FUNCTION for the MSVC compiler. Please check when you get a chance and let me know if it doesn’t work. I’m not suggesting you replicate that approach though, just use EPICS_FUNCTION if it’s available and define your own if not.

MarkRivers:
@mark0n and @anjohnson 
I just tried this on Linux and Windows.  I am using the EPICS_FUNCTION macro.

This is the line of code I am using:
```
    asynPrint(pasynUser, ASYN_TRACEIO_DRIVER, "EPICS_FUNCTION=%s\n", EPICS_FUNCTION);
```

This is what I see on Windows.  
```
2018/06/28 16:22:39.721 EPICS_FUNCTION=testAsynPortDriver::writeFloat64
2018/06/28 16:22:42.020 EPICS_FUNCTION=testAsynPortDriver::writeInt32
```
That is exactly what I want, the name of the function and nothing else.

However, this is what I see on Linux.
```
2018/06/28 16:23:11.959 EPICS_FUNCTION=virtual asynStatus testAsynPortDriver::writeFloat64(asynUser*, epicsFloat64)
2018/06/28 16:23:14.835 EPICS_FUNCTION=virtual asynStatus testAsynPortDriver::writeInt32(asynUser*, epicsInt32)
```
It prints out the full function definition, including the return type and all arguments.  This is much too verbose.  Is there a way to get just the function name, and not all of the other information?

MarkRivers:
I just found that on gcc EPICS_FUNCTION is defined as \_\_PRETTY_FUNCTION\_\_ which returns not just the function name but also the signature.  \_\_func\_\_ or \_\_FUNCTION\_\_ should return only the function name.  \_\_FUNCTION\_\_ is defined on both Linux and Windows so I tried that.

This is what I get on Windows, it is the same as EPICS_FUNCTION and is exactly what I want.
```
2018/06/28 16:42:21.008 __FUNCTION__=testAsynPortDriver::writeFloat64
2018/06/28 16:42:22.594 __FUNCTION__=testAsynPortDriver::writeInt32
```
However, this is what I get on Linux.  Note that it does not include the class name, only the function name.  So it seems that on Linux I either get too much or not enough?
```
2018/06/28 16:41:03.471 __FUNCTION__=writeInt32
2018/06/28 16:41:06.083 __FUNCTION__=writeFloat64
```

MarkRivers:
This page suggests extracting the class name from \_\_PRETTY_FUNCTION\_\_ at run-time on gcc.
https://stackoverflow.com/questions/1666802/is-there-a-class-macro-in-c

I guess there is no easy way to get just class::function in gcc.

ralphlange:
Well, when using overloaded functions in C++, I can see cases where I would want the full signature to be shown.
Admittedly that is true for only a few places, so being able to set the verbosity might be the best way to go.

PR #81, opened 2019-02-13
Fix output callback

### LINKED ISSUE #80: Deadlocks with asyn:READBACK on output records
There is a problem with output records with the info tag asyn:READBACK, i.e. output records that update their values from driver callbacks.

The interrupt callback function for output records calls dbScanLock because it is reading the PACT field of the record.  This can cause deadlocks because output records with synchronous drivers (ASYN_CANBLOCK=0) take the locks in the order 
1) dbScanLock (taken by EPICS base record processing)
2) driverLock (taken when device support process function calls the driver)

However, the callback function takes  the locks in the order 
1) driverLock (taken by the driver prior to the callback)
2) dbScanLock (taken in the callback function in device support)

Since the locks are taken in the opposite order deadlocks can occur.

Device support is looking at PACT to determine if the record is currently in the middle of
asynchronous record processing. 

A potential fix is to use a private variable to hold the information about whether asynchronous processing is in progress, so only the private mutex is needed, and not dbScanLock.

MarkRivers:
This pull request mostly fixes #80. 

The fix is to use a local variable to hold the information on whether asynchronous record processing is in progress, rather than using PACT.  This way only a local mutex, and not dbScanLock, can be used.

The fix has been applied devAsynInt32, devAsynOctet, and devAsynUInt32Digital.  It cannot be applied to devAsynOctet at this time.  This is because devAsynOctet copies the value directly to the record in the callback function if ring buffers are disabled, which is the default.  dbScanLock must therefore be called.  devAsynOctet output records with asyn:READBACK=1 and synchronous drivers may thus experience deadlocks.

### DIFF
--- a/asyn/devEpics/devAsynFloat64.c
+++ b/asyn/devEpics/devAsynFloat64.c
@@ -67,7 +67,7 @@ typedef struct devPvt{
     void              *float64Pvt;
     void              *registrarPvt;
     int               canBlock;
-    epicsMutexId      ringBufferLock;
+    epicsMutexId      devPvtLock;
     ringBufferElement *ringBuffer;
     int               ringHead;
     int               ringTail;
@@ -79,6 +79,7 @@ typedef struct devPvt{
     int               numAverage;
     int               isAiAverage;
     int               isIOIntrScan;
+    int               asyncProcessingActive;
     CALLBACK          processCallback;
     CALLBACK          outputCallback;
     int               newOutputCallbackValue;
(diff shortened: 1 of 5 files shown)


MarkRivers:
@anjohnson thanks for the suggestion, I have taken it.

I found that the asyn/testOutputCallback application could quickly generate the deadlock on the master branch when the driver was set to synchronous (ASYN_CANBLOCK=0) and little IDL or Python script quickly wrote to the PVs in that application.  Details are in a new comment in #80.  I added the IDL test script to the iocTestOutputCallback directory.

However, I found that the version in this PR also had the deadlock.  The problem was that although the record lock (dbScanLock) was no longer an issue there was a deadlock with pPvt->devPvtLock.  That lock needs to be released before calling queueRequest or we have a similar problem with synchronous drivers with the locks being taken in opposite order.  Once I fixed that then the testOutputCallback application no longer deadlocks.

PR #88, opened 2019-09-16
Add drvAsynFTDIPort

brunoseivam:
This PR adds a new driver to asyn, drvAsynFTDIPort, that communicates with an FTDI device (USB to Serial converter). The original code is from Philip Taylor (Observatory Sciences) and is reproduced here with permission, and with my modifications.


norumwe12:
I'm interested in learning more about the use case for this.  The operating systems of which I’m aware make FTDI serial ports accessible as standard termios lines which can be used with the ASYN serial port driver.

> On Sep 15, 2019, at 5:30 PM, Bruno Martins <notifications@github.com> wrote:
> 
> This PR adds a new driver to asyn, drvAsynFTDIPort, that communicates with an FTDI device (USB to Serial converter). The original code is from Philip Taylor (Observatory Sciences) and is reproduced here with permission, and with my modifications.
> 
> 
— 
Eric Norum
wenorum@lbl.gov

MarkRivers:
I was going to ask the same question. We have used devices like /dev/ttyUSB0. What does this driver do that one cannot do with such a device?

domitto:
Direct FTDI access gives much more access to the hardware which are not available using the legacy mode (/dev/ttyUSBx).
Like working in different modes: asynchronous, synchronous or bit bang mode. Writing to the eeprom chip, setting buffers size, latency, etc.
I'm working with the nPoint LC400 controller and the user manual recommends using FTDI access instead of VCP (legacy) for the fastest communication speed (3Mbps using FTDI).

Issue #90, opened 2019-10-09
devAsynXxx error message spamming


bfrk:
This is the same problem that I fixed some time ago for the asyn re-connect logic. Device support currently issues a message for every failed asyn read or write, in other words every time the record gets processed. Instead it should remember the last status and print a message only when the status changes. Below is a patch (relative tio asyn-4.35) that fixes this for devAsynInt32 and devAsynFloat64 but I think a similar change should be made to all device supports.

```
diff -rN -u old-4-35/asyn/devEpics/devAsynFloat64.c new-4-35/asyn/devEpics/devAsynFloat64.c
--- old-4-35/asyn/devEpics/devAsynFloat64.c	2019-10-09 18:38:58.736137891 +0200
+++ new-4-35/asyn/devEpics/devAsynFloat64.c	2019-10-09 18:38:58.736137891 +0200
@@ -74,6 +74,7 @@
     int               ringSize;
     int               ringBufferOverflows;
     ringBufferElement result;
+    asynStatus        lastStatus;
     epicsFloat64      sum;
     interruptCallbackFloat64 interruptCallback;
     int               numAverage;
@@ -333,10 +334,13 @@
         asynPrint(pasynUser, ASYN_TRACEIO_DEVICE,
             "%s %s::%s process value=%f\n",pr->name, driverName, functionName,pPvt->result.value);
     } else {
-        asynPrint(pasynUser, ASYN_TRACE_ERROR,
-              "%s %s::%s process read error %s\n",
-              pr->name, driverName, functionName,pasynUser->errorMessage);
+        if (pPvt->result.status  != pPvt->lastStatus) {
+            asynPrint(pasynUser, ASYN_TRACE_ERROR,
+                  "%s %s::%s process read error %s\n",
+                  pr->name, driverName, functionName, pasynUser->errorMessage);
+        }
     }
+    pPvt->lastStatus = pPvt->result.status;
     if(pr->pact) callbackRequestProcessCallback(&pPvt->processCallback,pr->prio,pr);
 }
 
@@ -354,10 +358,13 @@
         asynPrint(pasynUser, ASYN_TRACEIO_DEVICE,
             "%s %s::%s process val %f\n",pr->name, driverName, functionName,pPvt->result.value);
     } else {
-       asynPrint(pasynUser, ASYN_TRACE_ERROR,
-           "%s %s::%s pPvt->result.status=%d, process error %s\n",
[...]

PR #91, opened 2019-10-10
devEpics: process callbacks issue error only on status change

bfrk:
See [https://github.com/epics-modules/asyn/issues/90](https://github.com/epics-modules/asyn/issues/90). This is a cleaned up and extended version of the patch I attached to the issue.


MarkRivers:
I like this fix, thanks.  I was not seeing these messages because typically when my devices would return errors they were actually disconnected and messages in that case were already suppressed with reportQueueRequestStatus().

PR #96, opened 2020-01-03
Fix debug output by removing narrowing cast

mark0n:
See discussion here:
https://github.com/epics-modules/asyn/pull/85

### DIFF
--- a/asyn/miscellaneous/asynInterposeEos.c
+++ b/asyn/miscellaneous/asynInterposeEos.c
@@ -232,7 +232,8 @@ static asynStatus readIt(void *ppvt,asynUser *pasynUser,
              pasynUser,peosPvt->inBuf,peosPvt->inBufSize,&thisRead,&eom);
         if(status==asynSuccess) {
             asynPrintIO(pasynUser,ASYN_TRACEIO_FILTER,peosPvt->inBuf,thisRead,
-                "%s read %d bytes eom=%d\n",peosPvt->portName, (int)thisRead, eom);
+                "%s read %llu bytes eom=%d\n", peosPvt->portName,
+                (unsigned long long)thisRead, eom);
             /*
              * Read could have returned with ASYN_EOM_CNT set in eom because
              * the number of octets available exceeded inBufSize.  This is not


norumwe12:
Is the ‘z’ modifier supported by all environments in which ASYN is used?  I wonder about older vxWorks and RTEMS versions, in particular.   Is the (int) cast actually causing problems somewhere?  Would a cast to unsigned long with the corresponding format specified change to ‘%lu’ be adequate?

anjohnson:
VxWorks up to at least 6.9 does not support the `%z` modifier. I would favor using `%lu` and a cast to `unsigned long` myself, or `%llu` and `unsigned long long` if support for 64-bits is necessary (all our supported targets implement that as 64-bit).

mark0n:
Wow, this would be so much simpler with a C99 compatible compiler. The most portable way nowadays would probably be the following:
```
#include <inttypes.h>
asynPrintIO(pasynUser,ASYN_TRACEIO_FILTER,peosPvt->inBuf,thisRead, "%s read %"PRIu32" bytes eom=%d\n",peosPvt->portName, thisRead, eom);
```
But of course this also requires a C99-compatible compiler... So I have updated the PR to use `%llu` plus an ugly cast (which at least shouldn't be narrowing, though).

norumwe12:
Do older vxWorks and RTEMS implementations support ‘%llu’?  I think that casting to an unsigned long and using %lu would be adequate.  Are there really any real world applications using interposeEos with transfers exceeding 2^32-1 bytes?

MarkRivers:
As of asyn R4-37 asynDriver.h does the typedefs for asynInt64 and asynUInt64 on compilers where EPICS base does not do so.  Thus, I think you should cast size_t to epicsUInt64 and use %llu.  That should be more portable and less verbose than unsigned long long.

anjohnson:
VxWorks 5.5 did fully support `long long`. VxWorks 5.4 had compiler support for it too, but was missing some of the intrinsic routines needed by certain GCC-compiled constructs.

I agree that using a 64-bit value here seems somewhat excessive, but the rules for stdarg require a cast of some kind when passing a `size_t` in portable code.

MarkRivers:
> Do older vxWorks and RTEMS implementations support ‘%llu’?

I don't know about RTEMS.  However, I just committed a change that is like this PR but with (epicsUInt64) rather than (unsigned long long).  I checked and there were no compiler warnings or errors on the APS Jenkins build server with vxWorks 5.5.

For now I think we need to use this solution of cast to epicsUInt64 and print with %llu.  asyn certainly needs to work with vxWorks 6.9 which according to Andrew still does not support %z.

PR #97, opened 2020-01-03
SPI FTDI

jtagger:
Added minimal SPI capability for FTDI driver to enable AD9915 support.

### DIFF
--- a/asyn/drvAsynFTDI/drvAsynFTDIPort.cpp
+++ b/asyn/drvAsynFTDI/drvAsynFTDIPort.cpp
@@ -39,6 +39,7 @@ typedef struct {
     int               FTDIproduct;
     int               FTDIbaudrate;
     int               FTDIlatency;
+    int		      FTDImode;          /* UART = 0; SPI = 1 */
     char              *portName;
     FTDIDriver         *driver;
     unsigned long      nRead;
@@ -313,7 +314,12 @@ connectIt(void *drvPvt, asynUser *pasynUser)
     }
 
     // Create the driver
-    ftdi->driver = new FTDIDriver();
+
+    printf("connectIt: ftdi->FTDImode=%d - ", ftdi->FTDImode);
+    if( ftdi->FTDImode & 0x01 ) printf( "SPI\n");
+    else                        printf("UART\n");
+
+    ftdi->driver = new FTDIDriver(ftdi->FTDImode);
 
     // Set the vendor & product IDs
     ftdi->driver->setVPID(ftdi->FTDIvendor, ftdi->FTDIproduct);
@@ -511,8 +517,17 @@ drvAsynFTDIPortConfigure(const char *portName,
                          int latency,
                          unsigned int priority,
                          int noAutoConnect,
-                         int noProcessEos)
+                         int flags)
 {
+
+    int noProcessEos = ((flags & NO_PROC_EOS_BIT)==NO_PROC_EOS_BIT)?1:0;
+    int SPI = ((flags & UART_SPI_BIT)==UART_SPI_BIT)?1:0;
+
+    printf("drvAsynFTDIPortConfigure: latency=%d\n", latency);
+    printf("drvAsynFTDIPortConfigure: priorioy=%d\n", priority);
+    printf("drvAsynFTDIPortConfigure: noAutoconnect=%d\n", noAutoConnect);
+    printf("drvAsynFTDIPortConfigure: flags=%d\n", flags);
+
     ftdiController_t *ftdi;
     asynInterface *pasynInterface;
     asynStatus status;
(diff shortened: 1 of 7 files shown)


jtagger:
> There are some things that make me think you may have used Tab characters in the file, such as line 42 in drvAsynFTDIPort.cpp? Also line 91 in ftdiDriver.h looks strange?
> 
> Please make sure there are no tabs, only spaces.

Replaced the TAB with space, edited a line, removed some unused variables.

asyn/drvAsynFTDI/ftdiDriver.cpp
```diff
@@ -455,23 +560,48 @@ FTDIDriverStatus FTDIDriver::write(const unsigned char *buffer, int bufferSize,
   struct ftdi_transfer_control *tc = ftdi_write_data_submit (
     ftdi_, (unsigned char *) buffer, (int) bufferSize
   );
-  rc = ftdi_transfer_data_done (tc);
+  rc[8] = ftdi_transfer_data_done (tc);
```
brunoseivam: This looks like an out of bounds access. Should it be `rc[7]`?

asyn/drvAsynFTDI/ftdiDriver.cpp
```diff
@@ -455,23 +560,48 @@ FTDIDriverStatus FTDIDriver::write(const unsigned char *buffer, int bufferSize,
...
+    }
 #endif
 
   gettimeofday(&ctime, NULL);
   tnow = ((ctime.tv_sec - stime.tv_sec) * 1000) + (ctime.tv_usec / 1000);
   debugPrint("%s : Time taken for write => %ld ms\n", functionName, (tnow-(mtimeout-timeout)));
-  if (rc > 0){
-    debugPrint("%s : %d bytes written\n", functionName, rc);
-    *bytesWritten = (size_t)rc;
+  if (rc[3] > 0){
```
brunoseivam: It seems that throughout this function only `rc[3]` and `rc[7]` are ever used, and disjointly. Maybe I am missing the reason for using an array of return codes, but I believe it would be better to change these instances to simply use `rc` (a regular, singular `size_t`).

asyn/drvAsynFTDI/drvAsynFTDIPort.cpp
```diff
@@ -511,8 +517,17 @@ drvAsynFTDIPortConfigure(const char *portName,
                          int latency,
                          unsigned int priority,
                          int noAutoConnect,
-                         int noProcessEos)
+                         int flags)
 {
+
+    int noProcessEos = ((flags & NO_PROC_EOS_BIT)==NO_PROC_EOS_BIT)?1:0;
+    int SPI = ((flags & UART_SPI_BIT)==UART_SPI_BIT)?1:0;
```
brunoseivam: I understand repurposing the `noProcessEos` argument makes it easy to implement this feature, but I worry that this might be confusing for the user since other asynDrivers do not do that (AFAIK). However, I'd defer to @MarkRivers on this point. @jtagger , would it be possible to set the SPI/UART operation mode through the existing `setOption`/`getOption` interface? Or does it have to be set at creation time?

brunoseivam: Also, instead of `int SPI`, could you call this local variable `int mode` or something similar? Further below when you do `ftdi->FTDImode = SPI` I thought for a moment that you were defaulting to SPI mode.

jtagger:
Made some corrections: Programming AD9915 through SPI is very low level. Depending on where in the stream the binary bytes are sent (through the SPI) some bytes mean ‘data,’ some commands, (like ‘pull certain pin high or low’), some specify length etc. Unsure what will happen if to initialize to UART, then start setting Options. The SPI sequences are very fragile and susceptible getting out of sync easily. That is why I preferred to stick to pre-determined mode from the creation. Multiple return codes rc[8] - I used them at debugging for compiling one message from many codes. Debugging for every new chip will require them again.

brunoseivam:
Hi @jtagger , thanks for addressing our comments. Since multiple return codes are useful, would you mind adding a comment right above `int rc[16]` line explaining its purpose? It will be useful for future code reviewers.

Other than this, I am ready to approve this PR.

brunoseivam:
Looks good to me, unless @MarkRivers has any objections I can merge it.

jtagger:
In order to make use of the SPI interface to AD9915 (for basic RF amplitude and phase control) – it is necessary to connect 3 other wires (in addition to 3 SPI lines) of FTDI to AD9915. Furthermore, it is necessary to use a very specific stream protocol and configure some jumpers on the AD9915 board. (Other chips may need other configurations.) Do you think it appropriate to include some minimal setup description to ‘asynDriver.html / FTDI Port’ or it would be too much details?

MarkRivers:
I think it would be good to add that documentation.

PR #98, opened 2020-01-09
asynEpicsUtils.c: Remove compiler warnings about strncpy

tboegi:
The code snippet
  len = strlen(str);
  dst = malloc(len + 1);
  strncpy(dst, src, len);
  dst[len] = '\0';

Gives a compiler warning:
  strncpy’ output truncated before terminating nul copying as many bytes
  from a string as its length [-Wstringop-truncation

As the dst string has the exact right length, we don't need strncpy.
Use memcpy instead.

### DIFF
--- a/asyn/devEpics/asynEpicsUtils.c
+++ b/asyn/devEpics/asynEpicsUtils.c
@@ -86,10 +86,7 @@ static asynStatus parseLink(asynUser *pasynUser, DBLINK *plink,
         if(*p) {
             p = skipWhite(p,0);
             if(*p) {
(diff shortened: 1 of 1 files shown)


xiaoqiangwang:
From its functionality, isn't it doing the same as [epicsStrDup](https://github.com/epics-base/epics-base/blob/7.0/modules/libcom/src/misc/epicsString.c#L234), which is portable.

tboegi:
Good point:
Should I send a new version ?

ralphlange:
Just push to your branch.

PR #100, opened 2020-01-16
Make better use of Travis-CI by using the ci-scripts module

ralphlange:
Use ci-scripts v2.1.0 to blow up the Travis configuration to 19 different builds:
- Against EPICS 7.0 / 3.15.7 / 3.14.12.8 / 3.14.12.2
- Using g++ and clang
- Static and dynamic builds
- Cross-compilation to Windows using MinGW
- Cross-compilation to RTEMS 4.9 and 4.10 (including running the tests under qemu)
- On MacOS
- EPICS_LIBCOM_ONLY builds

Resulting [Travis builds will look like this](https://travis-ci.org/ralphlange/asyn/builds/637935310).

Feel free to change/add/remove jobs as you see fit. Happy to help with or review changes.


ralphlange:
The Travis RP build (in the "check" directly above) shows the worst-case build times when all dependencies (to be cached) have to be compiled from source.
Once the cache has been built, the regular build (link in the description) shows the best-case build times when only asyn itself has to be compiled.

ralphlange:
Wait - needs one more fix to run the tests on RTEMS...

ralphlange:
Ok, ready to go.
Thanks a lot to @mdavidsaver for fixing the RTEMS test - would have taken me ages.

MarkRivers:
I just pulled this and looked at the initial Travis build.  No problems, and it took 36 minutes because it had to build base and seq.  I then updated the RELEASE_NOTES.html to document adding ci-scripts, and the next Travis build only need to do asyn itself, and took 10 minutes.

This is great, thanks very much!

I'm now sold and have added an item to the Codeathon to use ci-scripts in lots of synApps modules.

PR #101, opened 2020-01-28
asynRecord.: Remove warning `...comparison of unsigned enum < 0...`

tboegi:
In asynExceptionToString() a range check is done to make sure that
we don't read outside asynExceptionStrings[].

The current code gives a warning, since modern compiler detects
that this very enum is always >= 0 and a comparison against < 0
does not make sense,

A more clean way of accessing an array is to use a variable of type size_t.
Cast the enum into size_t before doing the range check.
Use a help variable "idx" for this, the compiler will optimize it away.

### DIFF
--- a/asyn/asynRecord/asynRecord.c
+++ b/asyn/asynRecord/asynRecord.c
@@ -2064,7 +2064,7 @@ static void resetError(asynRecord * pasynRec)
 static const char * asynExceptionStrings[] = { ASYN_EXCEPTION_STRINGS };
 const char * asynExceptionToString( asynException e )
 {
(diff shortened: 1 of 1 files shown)


MarkRivers:
Why not just eliminate the test for < 0 and not bother with introducing idx?

tboegi:
enums can have values like -1, like "unkown", "undefined" or whatever.
That is why I want to have a cast into "(unsigned) size_t", which covers that case as well,
so that the test is future-proof for any definition of the enum.

tboegi:
Now idx is removed, thanks for the review.

PR #102, opened 2020-01-28
Remove the path when printing file/line

tboegi:
When the file/line printing is enabled by the bit ASYN_TRACEINFO_SOURCE,
file names are printed with help of the __FILE__ macro.
Due to our build system, the compiler is invoked with file names
like "../devMotorAsyn.c", and this is what we find in the IOC log file later.

However, the "../" is no useful information, so make it possible to
skip the "path-component" from file before the printing is done.

This will shorten the IOC log a little bit, and make it easier to read.

### DIFF
--- a/asyn/asynDriver/asynManager.c
+++ b/asyn/asynDriver/asynManager.c
@@ -472,6 +472,17 @@ static void reportConnectStatus(port *pport, portConnectStatus status, const cha
     }
 }
 
+static const char *asynStripPath(const char *file)
+{
+  const char *ret = strrchr(file, '/');
+  if (ret) return ret + 1;
+#if (defined(CYGWIN32) || defined(_WIN32))
+  ret = strrchr(file, '\\');
+  if (ret) return ret + 1;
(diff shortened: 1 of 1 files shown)


tboegi:
This is more like an RFC: Request For Comments,

MarkRivers:
This seems like it is introducing complexity in the user interface because of a build system specific to one site.  I would rather just always strip the path then have another ASYN_TRACEINFO option.

The strip is only looking for forward slash / characters.  What about Windows, will the "file" ever contain \characters? What happens when you build for Windows with your build system?

tboegi:
Thanks for the rewiew, @MarkRivers 
The new version should handle Windows as well, 
(both cygwin and Windows can use forward and backward slashes.)
The code change is 20% shorter now

tboegi:
@MarkRivers Coule you have another look at thos PR, please ?

MarkRivers:
I have merged this and tested on Linux and Windows.

PR #104, opened 2020-02-14
Add support for lsi, lso and printf records

FreddieAkeroyd:
Add support for lsi, lso and printf records to asynOctet.

TODO
 
* needs a clarification on string length to return i.e. whether to include terminating NULL in length. It looks like lsi/lso in epics base currently do include NULL in length.
* lsi/lso/printf records only became available with 3.15, so need to add conditionals to skip compilation on 3.14


FreddieAkeroyd:
Thanks - i have updated the todo list and will take a look at excluding 3.14. Also, do i need to add any tests?

MarkRivers:
I think asynCalc.dbd needs to be inside the conditional for CALC in the Makefile.

FreddieAkeroyd:
Thanks - yes that is better, no point installing something you will not be able to use.

MaticPogacnik:
Should we check if SSCAN is defined here: `https://github.com/epics-modules/asyn/blob/9c04876c063375a38a59e9692260f84aa56e153f/testEpicsApp/src/Makefile#L14` ?

As it's done here: `https://github.com/epics-modules/asyn/blob/9c04876c063375a38a59e9692260f84aa56e153f/asyn/Makefile#L196`

We have an environment where is CALC is defined but SSCAN is not. For this reason build fails if I don't explicitly undefine CALC for the asyn.

PR #109, opened 2020-04-20
Add new protocols tcp&, udp&, udp*& that enable SO_REUSEPORT

MarkRivers:
Add new protocols tcp&, udp&, udp*& that enable SO_REUSEPORT (or SO_REUSEADDR on Windows)

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPPort.c
@@ -85,6 +85,10 @@
 # endif
 #endif
 
+/* On most systems we use SO_REUSEPORT but RTEMS and Windows don't have SO_REUSEPORT, use SO_REUSEADDR instead */ 
+#if defined(__rtems__) || defined(_WIN32)
+# define USE_SO_REUSEADDR
+#endif
 
 /* This delay is needed in cleanup() else sockets are not always really closed cleanly */
 #define CLOSE_SOCKET_DELAY 0.02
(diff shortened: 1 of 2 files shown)


MarkRivers:
This PR failed in Travis CI building on RTEMS.  SO_REUSEPORT appears to not be supported on RTEMS.

@norumwe12 do you know how we can handle this?  On Windows SO_REUSEADDR seems to work.  Would that also work on RTEMS?

norumwe12:
My guess is that RTEMS is using a relatively old fork of the BSD network stack.

For now I think that just protecting the code in question with 
#ifdef SO_REUSEPORT
#endif
would be all that could be done.

anjohnson:
Hey Eric, weren't you responsible for adding the RTEMS network stack in the first place? They are now finally replacing it in the 5.x series...

I haven't looked too closely but I'm pretty sure the Base sources have been using SO_REUSEADDR when SO_REUSEPORT is not available, so it might have been working on RTEMS. I think there has been some recent work in the 7.0 tree related to this but I haven't been following the details too closely. There may even be an API in osiSock.h to do this for you, but my recollection may be somewhat confused between what's happened on the 3.15 and 7.0 branches and that work might not have even been released yet, so you'll need to check (just to be aware what might happen in future Base releases).

prjemian:
Ready to merge?

prjemian:
CI [fails](https://travis-ci.org/github/epics-modules/asyn/jobs/677413207) on RTEMS for this change: 

    ../../asyn/drvAsynSerial/drvAsynIPPort.c:463: error: 'SO_REUSEPORT' undeclared (first use in this function)

MarkRivers:
I have fixed the RTEMS build by using SO_REUSEADDR rather than SO_REUSEPORT, the same as on _WIN32.  It builds OK on Travis, but I don't have a way of testing if it actually works.

prjemian:
Thanks!

PR #111, opened 2020-04-28
Update Makefile

hanlet:
For cross-compiling, these libraries were missing for the ./asynPortDriver/unittest/Makefile


MarkRivers:
That pull request fails because the unit tests need to run successfully on base 3.14.12 where those libraries have different names.

This Makefile has already changed in a way that probably fixes this problem.  The version you are trying to merge has conflicts.  Please try the latest master branch and if there are still problems open a new issue or PR.

PR #114, opened 2020-05-11
Take connect timestamp after connection attempt

krisztianloki:
autoConnectDevice() has a built-in protection that forces at least 2
seconds between connection attempts. However, with the current implementation
the timestamp is taken _before_ starting the connection, so if the connection
attempt takes about 2 seconds (or more) to timeout then that 2 second
'do-not-connect' window has already expired (or about to expire) by the
time connectAttempt() returns. Better to take the timestamp _after_
connectAttempt() returns.


krisztianloki:
Hi @MarkRivers, could you please merge this?
Thanks!

PR #115, opened 2020-05-29
Miscellaneous fixes

ralphlange:
A few largely unrelated fixes that were found by static code analysis (sonar/cppcheck).

### DIFF
--- a/asyn/asynDriver/asynManager.c
+++ b/asyn/asynDriver/asynManager.c
@@ -1779,12 +1779,12 @@ static asynStatus queueLockPort(asynUser *pasynUser)
     asynStatus status = asynSuccess;
     double timeout;
 
-    asynPrint(pasynUser,ASYN_TRACE_FLOW, "%s asynManager::queueLockPort locking port\n", pport->portName);
     if(!pport) {
         epicsSnprintf(pasynUser->errorMessage,pasynUser->errorMessageSize,
                 "asynManager::queueLockPort not connected");
         return asynError;
     }
+    asynPrint(pasynUser,ASYN_TRACE_FLOW, "%s asynManager::queueLockPort locking port\n", pport->portName);
     if (pport->attributes & ASYN_CANBLOCK) {   /* Asynchronous driver */
         plockPortPvt = epicsThreadPrivateGet(pport->queueLockPortId);
         if (!plockPortPvt) {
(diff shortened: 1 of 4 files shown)


ralphlange:
I would call it a fix.
The content of `drvLinuxGpib.c` is the definition of whitespace damage. Really really awful.

Do you want me to add some trailing tabs to the line, too? (Didn't have them before, but would fit better in the context "style".)

ralphlange:
"Fixed" the whitespace "damage".

MarkRivers:
drvLinuxGpib.c is now a model citizen for whitespace.  Tabs gone, trailing white space gone, indentation consistent throughout.

I wrote a perl script to replace tabs with 4 spaces, and ran it on all .h, .c, and .cpp files in asyn.  I then manually checked that indentation was correct on the converted files and fixed any problems.

I wrote another perl script to remove trailing whitespace and ran it on all of the same files in asyn.

ralphlange:
Oh, Mark - that was honestly not my intention! Thanks a lot.
Dirk Zimoch did a similar job with the Base sources recently. Good to know that things are improving.

MarkRivers:
No problem, it’s been on my list of things to do for a while.

areaDetector is next. Some users have editors that automatically trim white space, and so pull requests show lots of changes that I don’t want to see. Most of the repositories I’m responsible for are already largely tab-free, but they do have trailing white space.

tboegi:
You guys have so much fun, when I am sleeping ;-)
For the TAB/"trailing white space" we have 2 different scripts in production,
both are running as part of ci when push to github bitbucket.
They are far away from being perfect, but have managed to keep the repos free from
"white space damage".
The python script can even be used locally to fix the damage.
And/or  it can be used as a "git pre-commit hook" to prevent "bad" commits.
(A pre-commit hook is not allowed to change the content of the worktree on disk.
 It is allowed to do checks, and stop the commit by using "exit 1" )

Anyway, it could be good to do the cleanup and keep the repos clean.

I am not sure what the "best" solution is.
shell scripts are most portable, I would say. python is gaining popularity,
but it could be, that pearl is the one that is available everywhere 
(Oldish Linux distros, Mac, Windows, cygwin, oherOS. Any thougths from anyone ?)
 
https://github.com/EuropeanSpallationSource/m-epics-ethercatmc/blob/master/checkws.sh

https://bitbucket.org/europeanspallationsource/tc_generic_structure/src/master/check_fix_white_space.py

PR #117, opened 2020-06-16
WIP - PVI

GDYendell:
Creates new asynParam and asynParamSet classes to store the parameter definitions separately from the driver class and updates asynPortDriver to have an asynParamSet and call createParam on all the parameters stored within it from child classes. There is a new asynPortDriver constructor that takes a asynParamSet and calls createParams.

For full context see: https://github.com/areaDetector/ADCore/issues/446

### DIFF
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -132,6 +132,7 @@ ifeq ($(EPICS_LIBCOM_ONLY),YES)
 endif
 INC += asynParamType.h
 INC += paramErrors.h
+INC += asynParamSet.h
 INC += asynPortDriver.h
 asyn_SRCS += paramVal.cpp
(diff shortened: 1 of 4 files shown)


anjohnson:
We probably don't want to add the .dls-build-master file to the official distribution.

MarkRivers:
This PR has a conflict. I believe this is coming because you removed trailing whitespace before issuing the PR, but subsequently I have removed trailing whitespace in all files on the master branch. This also makes it very difficult to see just what files you really changed.

Can you merge master into the dls-controls:pvi branch so these conflicts and apparent changes go away?

GDYendell:
Thanks for making the whitespace changes, that makes things easier. The diff should be clearer now.

PR #119, opened 2020-06-18
Makefile(s): Decouple CALC and SSCAN

tboegi:
The current Makefile(s) failed to build a version, where CALC was
defined, but not SSCAN.
Decouple those 2 in the Makefiles.


tboegi:
Already fixed in
https://github.com/epics-modules/asyn/commit/8573487454c20f6698f144ac68eb51ae68a22450

PR #120, opened 2020-07-22
Fix so reuseport

MarkRivers:
This fixes the problem that SO_REUSEPORT is not supported on older versions of Linux.  Previously we used SO_REUSEADDR instead of SO_REUSEPORT on RTEMS and Windows.  This changes the logic to use SO_REUSEADDR instead of SO_REUSEPORT whenever SO_REUSEPORT is undefined.


MarkRivers:
SO_REUSEPORT is not supported on Debian 7 (Wheezy) for example.  On that system /usr/include/asm-generic/socket.h contains this:
```
/* To add :#define SO_REUSEPORT 15 */
```

ralphlange:
It depends on the Linux kernel version. `SO_REUSEPORT` was [added in kernel release cycle 3.9](https://lwn.net/Articles/542629/).

Your fix looks fine.

PR #122, opened 2020-08-20
HiSLIP support seems work with the Stream Device support.

noboruatkek:
Although there are places for improvement, mostly for Python support, it seems work with the Stream Device .
Minimum information to test this software with EPICS/asy/StreamDevice is givend in README.md (or README.rst).
Python support also works with simple case, but need some improvement for support of async-channel communication, Error handling and SRQ interrupt. 

I will appreciate your comments and suggestions.


MarkRivers:
This PR fails to build on Travis and Appveyor.  These are some of the errors:
```
/usr/include/c++/5/bits/c++0x_warning.h:32:2: error: #error This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.

../../asyn/drvAsynUSBTMC/drvAsynUSBTMC.c:44:3: error: #error "You need to get a newer version of libsb-1.0 (16 at the very least)"
 # error "You need to get a newer version of libsb-1.0 (16 at the very least)"
```

norumwe12:
Could you disable USB TMC support for the offending target?  Or get a newer libusb installed?  Version 16 is over 7 years old now.

MarkRivers:
> Could you disable USB TMC support for the offending target?  Or get a newer libusb installed?  Version 16 is over 7 years old now.

@norumwe12 I don't understand that error.
- It does not happen on the master branch, the master branch builds fine.
- It only happens in this PR.  Why?

Note that the error message needs to be fix, it says "libsb" rather than "libusb".  That had me confused.  I'll fix it.

MarkRivers:
> - It only happens in this PR. Why?

I see why.  On the master branch CONFIG_SITE does not build drvAsynUSBTMC by default, but this PR does.

PR #123, opened 2020-08-31
Replace shareLib.h with ASYN_API

MarkRivers:
This PR converts asyn from using the traditional shareLib.h with epicsShareFunc, epicsShareExtern, epicsShareClass, etc. to using the mechanism now used in base.  Added asynAPI.h, modified asyn/Makefile,  and changed code to use the new mechanism.

### DIFF
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -38,6 +38,7 @@ ifeq ($(TIRPC),YES)
 endif
 
 SRC_DIRS += $(ASYN)/asynDriver
+INC += asynAPI.h
 INC += asynDriver.h
 INC += epicsInterruptibleSyscall.h
 asyn_SRCS += asynManager.c
@@ -353,4 +354,10 @@ else
 	$(PERL) $(TOOLS)/makeIncludeDbd.pl $(devEpics_DBD) $(notdir $@)
 	@$(MV) $(notdir $@) $@
 endif
+
+$(LIB_PREFIX)asyn$(LIB_SUFFIX): USR_CPPFLAGS += -DBUILDING_asyn_API
+ifeq ($(SHARED_LIBRARIES),YES)
+  $(SHRLIB_PREFIX)asyn$(SHRLIB_SUFFIX): USR_CPPFLAGS += -DBUILDING_asyn_API
+endif
+
 
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -355,9 +355,4 @@ else
 	@$(MV) $(notdir $@) $@
 endif
 
-$(LIB_PREFIX)asyn$(LIB_SUFFIX): USR_CPPFLAGS += -DBUILDING_asyn_API
-ifeq ($(SHARED_LIBRARIES),YES)
-  $(SHRLIB_PREFIX)asyn$(SHRLIB_SUFFIX): USR_CPPFLAGS += -DBUILDING_asyn_API
-endif
-
-
+USR_CPPFLAGS += -DBUILDING_asyn_API
--- a/.travis.yml
+++ b/.travis.yml
@@ -113,6 +113,15 @@ jobs:
 #  - env: SET=base7 BCFG=static-debug
 #    compiler: vs2017
 #    os: windows
+  - env: SET=base3-14
+    compiler: vs2017
+    os: windows
+  - env: SET=base3-14 BCFG=static
+    compiler: vs2017
+    os: windows
+  - env: SET=base3-14 BCFG=debug
+    compiler: vs2017
+    os: windows
 
 # Older Base releases
   - env: SET=base3-15
--- a/asyn/asynDriver/asynAPI.h
+++ b/asyn/asynDriver/asynAPI.h
@@ -1,0 +1,33 @@
+/* This is a generated file, do not edit! */
+
+#ifndef INC_asynAPI_H
+#define INC_asynAPI_H
+
+#if defined(_WIN32) || defined(__CYGWIN__)
+
+#  if !defined(epicsStdCall)
+#    define epicsStdCall __stdcall
+#  endif
+
+#  if defined(BUILDING_asyn_API) && defined(EPICS_BUILD_DLL)
+/* Building library as dll */
+#    define ASYN_API __declspec(dllexport)
+#  elif !defined(BUILDING_asyn_API) && defined(EPICS_CALL_DLL)
+/* Calling library in dll form */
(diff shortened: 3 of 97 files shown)


ralphlange:
Doesn't that break compatibilty with older base versions?
ASYN currently claims to work with base "3.14.12.2 or later".

asyn/Makefile
```diff
@@ -354,3 +355,9 @@ else
 	@$(MV) $(notdir $@) $@
 endif
 
+$(LIB_PREFIX)asyn$(LIB_SUFFIX): USR_CPPFLAGS += -DBUILDING_asyn_API
+ifeq ($(SHARED_LIBRARIES),YES)
+  $(SHRLIB_PREFIX)asyn$(SHRLIB_SUFFIX): USR_CPPFLAGS += -DBUILDING_asyn_API
+endif
```
mdavidsaver, line 361: Since libasyn is the only output from this directory, this could be simplified as:

```
USR_CPPFLAGS += -DBUILDING_asyn_API
```

mdavidsaver:
> Doesn't that break compatibilty with older base versions?

I don't think so.  (cf. travis-ci results)

MarkRivers:
I don’t think it using any new feature from base. I generated the asynAPI.h file manually using the Perl script from base 7.0.4.1. The new rules in the Makefile don’t seem to require any new features.

It passed the Travis build which includes building on 3.14.

anjohnson:
I’m on vacation this week, if the Appveyor builds succeed I would see no reason not to merge this, although I haven’t looked at it in detail.

MarkRivers:
Most of the Appveyor builds are working.  Two are failing:
- VS2019, dynamic build, base 3.14
- VS2019, dynamic-debug build, base 3.14

The failure is this:
```
link -nologo  -LTCG -incremental:no -opt:ref -release  -MACHINE:X64             
-out:asynPortDriverTest.exe        asynPortDriverTest.obj      
..\\..\\..\\..\\lib\\windows-x64\\asyn.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\recIoc.lib 
 C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\softDevIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\miscIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\rsrvIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\dbtoolsIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\asIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\dbIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\registryIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\dbStaticIoc.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\ca.lib  
C:\\Users\\appveyor\\.cache\\base-3.14\\lib\\windows-x64\\Com.lib   
asynPortDriverTest.obj : error LNK2001: unresolved external symbol "public: virtual __cdecl asynParamClient::~asynParamClient(void)" (??1asynParamClient@@UEAA@XZ)
asynPortDriverTest.obj : error LNK2001: unresolved external symbol "public: __cdecl asynParamClient::asynParamClient(char const *,int,char const *,char const *,double)" (??0asynParamClient@@QEAA@PEBDH00N@Z)
asynPortDriverTest.obj : error LNK2001: unresolved external symbol pasynInt32SyncIO
asynPortDriverTest.obj : error LNK2001: unresolved external symbol "public: __cdecl asynPortDriver::asynPortDriver(char const *,int,int,int,int,int,int,int)" (??0asynPortDriver@@QEAA@PEBDHHHHHHH@Z)
asynPortDriverTest.exe : fatal error LNK1120: 4 unresolved externals
```

The failure is only for:
- Base 3.14, not 3.15 or 7.0
- Dynamic builds, not static builds

Any ideas?

tboegi:
Somewhat shooting into the blue, is there a mismatch between the .c and the .h file ?


asyn/interfaces/asynInt32SyncIO.h 
ASYN_API extern asynInt32SyncIO *pasynInt32SyncIO;

 asyn/interfaces/asynInt32SyncIO.c 
asynInt32SyncIO *pasynInt32SyncIO = &interface;

mdavidsaver:
Comparing with the same build for 3.15 I see that in the failed build asynPortDriverTest.cpp is not built with `-DEPICS_BUILD_DLL -DEPICS_CALL_DLL`.  Looking back, I see that the Makefile logic was reversed back then, and instead static builds pass `-DEPICS_DLL_NO`.

https://github.com/epics-base/epics-base/blob/3.14/src/libCom/misc/shareLib.h

I think this could be resolved by prepending something like this to asynAPI.h

```c
#include <epicsVersion.h>

#ifndef VERSION_INT
//! Construct version number constant.
#  define VERSION_INT(V,R,M,P) ( ((V)<<24) | ((R)<<16) | ((M)<<8) | (P))
#endif

#ifndef EPICS_VERSION_INT
#  define EPICS_VERSION_INT VERSION_INT(EPICS_VERSION, EPICS_REVISION, EPICS_MODIFICATION, EPICS_PATCH_LEVEL)
#endif

#if defined(_WIN32) && EPICS_VERSION_INT<VERSION_INT(3,14,0,0) && !defined(EPICS_DLL_NO)
#    define EPICS_BUILD_DLL
#    define EPICS_CALL_DLL
#endif
```

MarkRivers:
@mdavidsaver your diagnosis seems correct.  I am going to try to reproduce the problem on Travis, since then testing the solution only takes 30 minutes instead of many hours.

mdavidsaver:
>  I am going to try to reproduce the problem on Travis

That would be ideal.  My fallback is to temporarily change the order, or remove jobs, from the appveyor config.

MarkRivers:
I added 3 more builds on Travis.  These all use base 3.14 and VS2017:
- dynamic
- static
- dynamic-debug.  

The dynamic builds both failed, as they did on Appveyor.

I then added the changes @mdavidsaver suggested for asynAPI.h, and those 3 builds all succeeded.  The RTEMS build failed, but that appears to be a stalled job, unrelated problem.

PR #125, opened 2020-09-26
Eliminate use of std::auto_ptr

MarkRivers:
@mdavidsaver in 2014 you improved asynPortDriver by using STL classes.  One of the changes you made was to use the std::auto_ptr class in one place.  std::auto_ptr is now deprecated and will be removed in C++ 17.  I therefore am proposing to remove it in this PR.

Can you comment on this?  What was gained by using std::auto_ptr over this code? Is there a better alternative that will work with older versions of C++?

### DIFF
--- a/asyn/asynPortDriver/asynPortDriver.cpp
+++ b/asyn/asynPortDriver/asynPortDriver.cpp
@@ -129,11 +129,10 @@ asynStatus paramList::createParam(const char *name, asynParamType type, int *ind
 
     if (this->findParam(name, index) == asynSuccess) return asynParamAlreadyExists;
 
-    std::auto_ptr<paramVal> param(new paramVal(name, type));
+    paramVal *param = new paramVal(name, type);
 
-    vals.push_back(param.get());
+    vals.push_back(param);
     flags.reserve(vals.size());
-    param.release();
     *index = (int)vals.size()-1;
     return asynSuccess;
 }


ralphlange:
auto_ptr has been deprecated in favor of unique_ptr and shared_ptr - depending on use case.

Are you planning on using C++17 any time soon? (EPICS Base is not at this point.)

MarkRivers:
> Are you planning on using C++17 any time soon? (EPICS Base is not at this point.)

No, asyn must be able to build on much older compilers.

What is the advantage of using a smart pointer in that code?

mdavidsaver:
>  std::auto_ptr is now deprecated and will be removed in C++ 17

I've something of a grudge again those who decided this.  It is all well and good to move forward, but in the case of auto_ptr it is just as likely to have the opposite effect since not everyone can/will require >=c++11, even in 2020.

In my projects, I avoid this warning by conditionally substituting `std::unique_ptr`:

https://github.com/epics-modules/mrfioc2/blob/0a62d1393779cd382dc8d68ec03d944727fc3686/mrfCommon/src/mrfCommon.h#L55-L65

Whether this is worthwhile for a single usage is something for @MarkRivers to decide.

> What is the advantage of using a smart pointer in that code?

To avoid the possibility of a memory leak if subsequent code throws an exception.  Specifically, `std::vector<T>::push_back()` or `std::vector<T>::reserve()` may do a memory allocation, which like any `malloc()`, may fail.  This throws `std::bad_alloc`.

mark0n:
I would like to remind you that we extensively discussed the topic of moving forward to C++11 at the EPICS meeting at ICALEPCS 2019. The suggestion to raise the bar to C++11 for EPICS Base 7.1 was received quite favorably by the community and I'm pretty sure the same would be true for Asyn as well. What I learned from our discussions is that moving to C++11 shouldn't be a big issue while C++14 might be a different story. Another key takeaway was that those of us who (for whatever reason) are stuck with an ancient compiler usually do not have a need to run the latest and greatest EPICS/Asyn code anyway. With that in mind I believe raising the bar to C++11 for Asyn wouldn't have any negative consequences. I actually recall at least [one place in the code where C++11-support would make Asyn code simpler](https://github.com/epics-modules/asyn/commit/851b6e7e95eaedadd6710c132eed513c1dfbca94) but I wouldn't be surprised if we would find range-base for loops, lambdas, `std::move` etc. useful as well.

In case you don't want to make this step at this time: Have you considered setting `USR_CXXFLAGS+=-std=c++03`? If you specify this even the most modern compilers should happily compile Asyn. Setting the C++ version might be a good idea anyway since it ensures developers get an error when they accidentally use modern C++ features in this code base. Plus it suppresses the annoying deprecation warning we can't do anything about as long as we need to support pre-C++11 compilers.

MarkRivers:
> I would like to remind you that we extensively discussed the topic of moving forward to C++11 at the EPICS meeting at ICALEPCS 2019. The suggestion to raise the bar to C++11 for EPICS Base 7.1 was received quite favorably by the community and I'm pretty sure the same would be true for Asyn as well.

@mark0n thanks for reminding me of that.  I just pushed to a new branch,  replace_autoptr_uniqueptr which simply replaces auto_ptr with unique_ptr.  I also added this line to the asyn/asyn/Makefile
```
USR_CXXFLAGS_Linux += -std=c++11
```
With that I was able to build on Centos7 (gcc 4.8.5) and Windows with Visual Studio 2015.

However, it will not build on vxWorks 6.9.4, which does not support C++11.  C++11 is only supported in vxWorks 7 and higher.  vxWorks 6.9.4 is heavily used at the APS on both the beamlines and the accelerator.  I don't think we can move to C++11 until vxWorks 7 is available for use at the APS. I don't know what Andrew's plans are for that.

I will create a PR for this new branch and see if there are other problems with the Travis and Appveyor builds.

mark0n:
@MarkRivers I would recommend to go with the same standard version for all targets. Have you tried `USR_CXXFLAGS+=-std=c++03`? This should prevent compilers that default to C++17 to fail when encountering `std::auto_ptr`.

mdavidsaver:
>  I also added this line to the asyn/asyn/Makefile
> 
> ```
> USR_CXXFLAGS_Linux += -std=c++11
> ```

Be aware that `-std=*` can [effect ABI](https://gcc.gnu.org/onlinedocs/libstdc++/manual/using_dual_abi.html).  I've run into this in the past when passing `std::string` as arguments.  The result was a linking error as the mangled symbol names differed.

PR #126, opened 2020-09-27
Replace autoptr uniqueptr

MarkRivers:
Requires C++11.

### DIFF
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -24,6 +24,7 @@ ifeq (vxWorks,$(findstring vxWorks, $(T_A)))
   CODE_CXXFLAGS=
 endif
 USR_CXXFLAGS_solaris-sparc=-instances=global
+USR_CXXFLAGS_Linux += -std=c++11
 
 LIBRARY_IOC += asyn
 
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -24,7 +24,10 @@ ifeq (vxWorks,$(findstring vxWorks, $(T_A)))
   CODE_CXXFLAGS=
 endif
 USR_CXXFLAGS_solaris-sparc=-instances=global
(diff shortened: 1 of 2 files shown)


MarkRivers:
The first Travis build failed because it was not setting the --std=c++11 flag on RTEMS.  I change the Makefile so that it sets that flag for all GNU builds.

MarkRivers:
I just tested this branch on RHEL6 (gcc 4.4.7).

It does not work.  

The -std=c++11 flag is not recognized:
```
cc1plus: error: unrecognized command line option "-std=c++11"
```

If I remove that flag I get an error on unique_ptr:
```
../../asyn/asynPortDriver/asynPortDriver.cpp:132: error: 'unique_ptr' is not a member of 'std'
```

Unfortunately RHEL6 is very important for beamlines.  Dectris is still shipping new Pilatus detectors with RHEL6 servers. These new Pilatus detectors are a critical part of our experiments, and I don't want to prevent using the latest versions of asyn and areaDetector on them.  Hopefully we can convince Dectris to move to RHEL 8 (or at least 7) but I don't know when they will do that.

The marCCD detector driver also only works on RHEL6.  

@mark0n so both vxWorks 6.9.4 and RHEL6 are preventing us from moving to C++11.

MarkRivers:
I just checked the Travis builds for RTEMS 4.10.  It fails the same way as RHEL6.  It has this error if the -std=c++11 flag is used:
```
cc1plus: error: unrecognized command line option "-std=c++11"
```
or this error if the flag is not used:
```
../../asyn/asynPortDriver/asynPortDriver.cpp:132: error: 'unique_ptr' is not a member of 'std'
```

MarkRivers:
This is not practical at the present time.

PR #127, opened 2020-09-27
Make std::auto_ptr work with modern compilers

mark0n:
Better solution for #125.

### DIFF
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -25,6 +25,8 @@ ifeq (vxWorks,$(findstring vxWorks, $(T_A)))
 endif
 USR_CXXFLAGS_solaris-sparc=-instances=global
 
+USR_CXXFLAGS+=-std=c++03
+
 LIBRARY_IOC += asyn
 
 asyn_SYS_LIBS_WIN32 = ws2_32 winmm
--- a/asyn/asynPortDriver/asynPortDriver.cpp
+++ b/asyn/asynPortDriver/asynPortDriver.cpp
@@ -129,10 +129,11 @@ asynStatus paramList::createParam(const char *name, asynParamType type, int *ind
 
     if (this->findParam(name, index) == asynSuccess) return asynParamAlreadyExists;
 
-    paramVal *param = new paramVal(name, type);
+    std::auto_ptr<paramVal> param(new paramVal(name, type));
 
-    vals.push_back(param);
+    vals.push_back(param.get());
     flags.reserve(vals.size());
+    param.release();
     *index = (int)vals.size()-1;
     return asynSuccess;
 }


MarkRivers:
This PR fails on RTEMS:
```
cc1plus: error: unrecognized command line option "-std=c++03"
```

It also fails on RHEL6:
```
cc1plus: error: unrecognized command line option "-std=c++03"
```

MarkRivers:
This PR also fails in vxWorks 6.9.4:
```
cc1plus: error: unrecognized command line option "-std=c++03"
```

MarkRivers:
I think this PR may be moot.  I have fixed the original warning by simply removing the std::auto_ptr from asynPortDriver.  Its only purpose was to prevent a memory leak if the allocation of a few bytes in that function failed.  As a practical matter if that allocation of a few bytes fails, then a memory leak is almost certainly the least of the problems, something is seriously amiss.

If we really want to guard against an exception generating a memory leak we can do it it in a try()/catch() in that function.

The fix for vxWorks does not fix RHEL6, RTEMS 4.10 and other older compilers.  Those older compilers don't support either std::unique_ptr() or -std=c++03.

mark0n:
I agree that in this particular case the simplest solution might be to go back to raw pointers. However, with new C++ versions being released every 3 years and EPICS Base and other support modules being affected by similar problems it might be worth finding a better solution than moving towards _older_ C++ code (the recommended path forward is actually to replace `std::auto_ptr` by C++11's `std::uniq_ptr`).

I have updated this PR with special handling for VxWorks and RTEMS - this is good enough to pass the Travis build but it's not ideal since it would break for people compiling for VxWorks 7 with a compiler which defaults to C++17.

I remember C++03 being our smallest common denominator. I'm wondering if this assumption is wrong or if the affected compilers are just lacking the `-std=c++03` command-line parameter. @MarkRivers could you please check the documentation of your compiler?

If only the command-line parameter is missing we could also test if the compiler understands the `-std=c++03` flag:
```
ifeq ($(shell $$COMPILER -c -x c++ -std=c++03 - < /dev/null > /dev/null; echo $$?), 0)
  USR_CXXFLAGS+=-std=c++03
endif
```
Not exactly pretty but maybe acceptable as a stop-gap solution until these ancient compilers have been retired? If that's the route we want to go I might need some advice on how to populate the `COMPILER` variable. @anjohnson, @mdavidsaver any thoughts?

MarkRivers:
Since the -std flag is a GNU thing then it seems like we should be testing the GNU version to see if that flag is supported?

On RHEL6 the gcc version is:
```
gcc (GCC) 4.4.7 20120313 (Red Hat 4.4.7-23)
```
and these are the allowed values for -std=
```

           c89
           iso9899:1990
               Support all ISO C90 programs (certain GNU extensions that conflict with ISO C90 are disabled). Same
               as -ansi for C code.

           iso9899:199409
               ISO C90 as modified in amendment 1.

           c99
           c9x
           iso9899:1999
           iso9899:199x
               ISO C99.  Note that this standard is not yet fully supported; see
               <http://gcc.gnu.org/gcc-4.4/c99status.html> for more information.  The names c9x and iso9899:199x
               are deprecated.

           gnu89
               GNU dialect of ISO C90 (including some C99 features). This is the default for C code.

           gnu99
           gnu9x
               GNU dialect of ISO C99.  When ISO C99 is fully implemented in GCC, this will become the default.
               The name gnu9x is deprecated.

           c++98
               The 1998 ISO C++ standard plus amendments. Same as -ansi for C++ code.

           gnu++98
               GNU dialect of -std=c++98.  This is the default for C++ code.

           c++0x
               The working draft of the upcoming ISO C++0x standard. This option enables experimental features
               that are likely to be included in C++0x. The working draft is constantly changing, and any feature
               that is enabled by this flag may be removed from future versions of GCC if it is not part of the
               C++0x standard.

           gnu++0x
               GNU dialect of -std=c++0x. This option enables experimental features that may be removed in future
               versions of GCC.
```

MarkRivers:
So there is a c++0x option for -std, but not c++03.

anjohnson:
The vxWorks-6.9 compiler (gcc-4.3.3) accepts the `-std=c++0x` flag too.

anjohnson:
However sites like PSI may want to build newer versions of Asyn with older versions of Base and using older VxWorks compilers that don't even have that flag, so I really don't think individual modules that aim to support older installations should be setting this. Also the test here for RTEMS will be broken very soon, RTEMS-5.1 comes with gcc-7.5.0.

We could look at adding a Makefile variable to the EPICS build system that specifies the minimum C++ version required to compile the code built by this Makefile. It would have to be optional since older Base versions wouldn't recognize it at all, and it could complicate the target configuration somewhat, but it might allow the build to apply the flag(s) appropriate for the latest version of the standard that this target's compiler supports. I think that's what @mark0n is really looking for. That suggestion is off-topic for this particular PR though, which I think is probably dead.

MarkRivers:
This issue is now moot since std::autoptr has been removed from asyn.

PR #128, opened 2020-10-07
remove warning implicit conversion from int to char

tboegi:
According to the C-standard a 'char' can either be signed,
reaching from -128..+127, or unsigned, reaching from 0..255

Some compilers spot out a warning like this:
warning: implicit conversion from 'int' to 'char'
      changes value from 255 to -1 [-Wconstant-conversion]

Remove the warning by adding an expliicit cast to char.


ralphlange:
Just FYI: different compilers have different signedness for `char` (without explicit specification), which can sometimes lead to more fun for portable code.
(See e.g. https://www.linuxtopia.org/online_books/an_introduction_to_gcc/gccintro_71.html)

PR #131, opened 2021-03-07
Make O.* match directory only

FreddieAkeroyd:


ralphlange:
Looks benign. What's the motivation?

FreddieAkeroyd:
I was going to link it as an example and if somebody was going to use it as the basis/update of a `.gitignore` then adding the `/` was just slightly more consistent. There is no actual problem I am aware of, and it is probably unlikely somebody would try and create a file called `o.something`

PR #133, opened 2021-05-28
Output records, asyn:READBACK=1 and late connect

tboegi:
When an output record is setup with PINI=0 and no VAL is specified
and asyn:READBACK=1 is set, the record will read the value from
the device and put that value into the record.

This works well when the driver has been able to read the value from the
hardware before the record in initialized.

When the driver reads the value (much) later, the normal callback/process
code will put the value into the record, but the record stays in the
"INVALID DRIVER UDF" alarm state (with UDF = 0).

Another process() will trigger the alarm processing (?) and
the record leaves the alarm state.
Note that this process() is only triggered when a new value is read from
the driver, or the VAL field of the record changes, so it may never happen.

Fix this issue for the 32 bit integer records.
Note: int64 and double64 need this fix as well.

### DIFF
--- a/asyn/devEpics/devAsynFloat64.c
+++ b/asyn/devEpics/devAsynFloat64.c
@@ -664,7 +664,10 @@ static long processAo(aoRecord *pr)
             if (pr->aslo != 0.0) val64 *= pr->aslo;
             val64 += pr->aoff;
             pr->val = val64;
-            pr->udf = 0;
+            if (pr->udf) {
+                pr->udf = 0;
+                recGblResetAlarms(pr);
+            }
         }
     } else if(pr->pact == 0) {
         /* ASLO/AOFF conversion */
--- a/asyn/devEpics/devAsynInt32.c
+++ b/asyn/devEpics/devAsynInt32.c
@@ -987,7 +987,10 @@ static long processAo(aoRecord *pr)
         /* We got a callback from the driver */
         if (pPvt->result.status == asynSuccess) {
             pr->rval = pPvt->result.value;
-            pr->udf = 0;
+            if (pr->udf) {
+                pr->udf = 0;
+                recGblResetAlarms(pr);
+            }
             value = (double)pr->rval + (double)pr->roff;
             if(pr->aslo!=0.0) value *= pr->aslo;
             value += pr->aoff;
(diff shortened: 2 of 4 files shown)


MarkRivers:
I disagree with this change.  I think it is just hiding the real problem.

If your poller has not yet run when the initBo initial read is done then the value will not have been set in the parameter library. 
 pasynInt32SyncIO->read() will return status=asynError and the record value will not be initialized.

When the poller does run after iocInit it will call setIntegerParam and callParamCallbacks.  That will cause the record to process, and udf will be set to 0 and STAT and SEVR set to NO_ALARM.

Please explain why this is not happening.

You have not posted your database or your driver, so I cannot look for errors in your code.

tboegi:
Hej Mark,


On 5/28/21 2:20 PM, Mark Rivers wrote:
> I disagree with this change. I think it is just hiding the real problem.
> 
> If your poller has not yet run when the initBo initial read is done then 
> the value will not have been set in the parameter library.
> pasynInt32SyncIO->read() will return status=asynError and the record 
> value will not be initialized.

Strictly speaking pasynInt32SyncIO->read() returns 10, which is
asynParamUndefined
(But that is just not important here)

> 
> When the poller does run after iocInit it will call setIntegerParam and 
> callParamCallbacks. That will cause the record to process, and udf will 
> be set to 0 and STAT and SEVR set to NO_ALARM.
> 
> Please explain why this is not happening.
That is my originally question...
asyn just sets UDF, but does not touch STAT or SEVR

What asyn does, is to call
(void)recGblSetSevr(pr, pPvt->result.alarmStatus, 
pPvt->result.alarmSeverity);

But that call does only raise STAT or SEVR, and does not lower them.
It could be (must be so), that base is resetting them, but I haven't 
found the code yet.


> 
> You have not posted your database or your driver, so I cannot look for 
> errors in your code.

Good point.

The master of
https://github.com/tboegi/m-epics-EthercatMC/
has all the driver, a simulator...

The Clutch is here:
<https://github.com/tboegi/m-epics-EthercatMC/blob/master/ethercatmcApp/Db/ethercatmcOpenClutch.template>

The important part of the driver is here:
<https://github.com/tboegi/m-epics-EthercatMC/blob/master/ethercatmcApp/src/ethercatmcIndexer.cpp#L1613-L1642>

It is quite "easy" to reproduce the problem.
Start the IOC and after it has been started, start the simulator.
My plan was to run this through gdb today,
but got stuck with other stuff.

Mark, your help is much appreciated.
If you want to look for mistakes in my code: good.
It is not so complex to set up a test system,
if you want, I can help you with that.

BR
/Torsten

PR #134, opened 2021-05-28
Add checkws.sh and fix trailing whitespace

tboegi:
Some trailing whitespace have sneaked in - remove them.
Add a checkws.sh script, that can be run manually and later integrated
into a ci script/action.


MarkRivers:
I fixed these files, but was not sure about the script.

PR #140, opened 2021-09-24
remove `using std::logic_error` from header files

krmpotic:


MarkRivers:
Can you explain the motivation for this change?

krmpotic:
I thought `using` in header files is bad practice, since all files including the header will also get influenced, probably unintentionally.

anjohnson:
I agree with the reasoning for this change. It should be documented in the Release Notes though, and since this may break downstream modules that build fine with earlier versions I would include examples of the errors to expect from at least one of our supported compilers and how to fix them in the downstream code.

PR #145, opened 2021-10-16
Base asynInterpose implementation

darcato:
Hello,
this adds a base implementation of an ``asynInterpose``, which does _absolutely nothing_. All the requests are passed directly to the underlying port.

This is useful as a guide for new users to implement a new interpose. One can just start from this base implementation and change only the functions that are useful for his use-case. Since I had to implement a custom ``asynInterpose``, this would have saved me a few hours.

I wrote this by looking at the other ``asynInterpose`` implementations but I am not an expert in asyn, so let me know if I missed something.

### DIFF
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -102,7 +102,7 @@ DBD += asyn.dbd
 INC += asynShellCommands.h
 INC += asynInterposeCom.h
 INC += asynInterposeEos.h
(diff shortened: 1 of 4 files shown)


norumwe12:
Very useful.  Would the name 'asynInterposeTemplate' perhaps be more descriptive?

PR #148, opened 2022-03-10
allow to specify an interface to bind the ip server port

lrossa:
Use the 2nd parameter of "drvAsynIPServerPortConfigure" for more than specifying a tcp/udp port only. The port (after the colon) works as before, the part before the colon is to select an interface (a host name or ip address). If the name cannot be resolved or is not an ip address, it will use the previous/old behavior and uses any interface. The example "localhost:port" will bind to the loopback interface on most machines only. To get the previous behavior, use an empty host ":port" or the any address "0.0.0.0:port".

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPServerPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPServerPort.c
@@ -57,6 +57,7 @@ typedef struct {
     unsigned int       portNumber;
     char              *portName;
     char              *serverInfo;
+    char              *interfaceInfo;
     int                maxClients;
     int                socketType;
     int                priority;
@@ -382,6 +383,7 @@ int createServerSocket(ttyController_t *tty) {
     int i;
     struct sockaddr_in serverAddr;
     int oneVal=1;
+    char srvaddrtxt[256];
     assert(tty);
     /*
      * Create the socket
@@ -395,6 +397,25 @@ int createServerSocket(ttyController_t *tty) {
         serverAddr.sin_family = AF_INET;
         serverAddr.sin_addr.s_addr = INADDR_ANY;
         serverAddr.sin_port = htons(tty->portNumber);
+
+        /*
+         * Got host or interface name, resolve it unless we see "localhost":
+         * This may be used in existing start scripts and we don't want to
+         * break them. If you need the loopback, use "127.0.0.1".
+         */
+        if (tty->interfaceInfo != NULL && *tty->interfaceInfo &&
+            epicsStrCaseCmp(tty->interfaceInfo, "localhost")) {
+            if (hostToIPAddr(tty->interfaceInfo, &serverAddr.sin_addr) < 0) {
+                printf("using ANY interface, cannot lookup '%s': %s\n",
+                       tty->interfaceInfo, strerror(SOCKERRNO));
+                return -1;
+          }
+        }
+
+        if (ipAddrToDottedIP(&serverAddr, &srvaddrtxt[0], sizeof(srvaddrtxt)) > 0) {
+            srvaddrtxt[sizeof(srvaddrtxt) - 1] = '\0';
+            printf("serverAddr: %s\n", srvaddrtxt);
+        }
         printf("serverPort: %i\n", tty->portNumber);
         if (tty->socketType == SOCK_DGRAM) {
             /* For Port reuse, multiple IOCs */
@@ -479,6 +500,8 @@ static void ttyCleanup(void *pPvt)
         epicsSocketDestroy(tty->fd);
     }
     free(tty->portName);
+    free(tty->serverInfo);
+    free(tty->interfaceInfo);
     free(tty);
 }
 
@@ -551,6 +574,7 @@ int drvAsynIPServerPortConfigure(const char *portName,
         return -1;
     }
     *cp = '\0';
+    tty->interfaceInfo = epicsStrDup(serverInfo);
 
     if ((protocol[0] == '\0')
             || (epicsStrCaseCmp(protocol, "tcp") == 0)) {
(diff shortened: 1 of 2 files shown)


asyn/drvAsynSerial/drvAsynIPServerPort.c
```diff
@@ -382,6 +383,7 @@ int createServerSocket(ttyController_t *tty) {
     int i;
     struct sockaddr_in serverAddr;
     int oneVal=1;
+    char srvaddrtxt[50];
```
tboegi, line 386: Is there a special reason to use 50 here ?
For IPv4 addresses we typically use 32.
If we want to prepared for IPv6, 64 is a typical value

documentation/asynDriver.html
```diff
@@ -5778,7 +5778,12 @@ <h3 id="drvAsynIPServerPort">
   <ul>
     <li>portName - The portName that is registered with asynManager.</li>
     <li>serverInfo - The Internet host name and port number to listen for connections
-      on (e.g. "localhost:4002" for TCP, "localhost:4002 UDP" for UDP).</li>
+      on (e.g. "localhost:4002" for TCP, "localhost:4002 UDP" for UDP), which will
+      bind only to the loopback interface on many machines. This allows to specify a
+      host name or ip address for multi-homed machines. If you want to specify <b>*any*</b>
+      interface, use an empty host <pre>":port"</pre> or the any address <pre>"0.0.0.0:port"</pre>
+      . The address has to be an existing interface on the host or if the name cannot
+      be resolved, it will use the any address like in the past.</li>
```
tboegi, line 5786: I think this may break existing scripts which rely on localhost:4002 to not do any binding to an IP address.
See my other comment in the code.
If you really want to bind to 127.0.0.1:4002, specify "127.0.0.1:4002" instead of "localhost:4002"

asyn/drvAsynSerial/drvAsynIPServerPort.c
```diff
@@ -395,6 +397,16 @@ int createServerSocket(ttyController_t *tty) {
         serverAddr.sin_family = AF_INET;
         serverAddr.sin_addr.s_addr = INADDR_ANY;
         serverAddr.sin_port = htons(tty->portNumber);
+
+        /* got host or interface name, resolve it */
+        if (tty->interfaceInfo != NULL && *tty->interfaceInfo)
+            if (hostToIPAddr(tty->interfaceInfo, &serverAddr.sin_addr) < 0)
+                serverAddr.sin_addr.s_addr = INADDR_ANY;
+
```
tboegi, line 405: There are 2 possible improvements:
- Keep "localhost:4002" as is to be backwards compatible
- Fail if the interfaceInfo can not be resolved.
```

        /*
         * got host or interface name, resolve it
         * unless we see "localhost":
         * This may be used be used in existing start scripts
         * and we don't want to break them
         */
        if (tty->interfaceInfo != NULL && *tty->interfaceInfo &&
            strcmp(tty->interfaceInfo, "localhost")) {
            if (hostToIPAddr(tty->interfaceInfo, &serverAddr.sin_addr) < 0) {
                //serverAddr.sin_addr.s_addr = INADDR_ANY;
                printf("Can't lookup '%s' : %s",
                       tty->interfaceInfo, strerror(SOCKERRNO));
                return -1;
            }
        }

        if (ipAddrToDottedIP(&serverAddr, &srvaddrtxt[0], sizeof(srvaddrtxt)) > 0) {
            srvaddrtxt[sizeof(srvaddrtxt) - 1] = '\0';
            printf("serverAddr: %s\n", srvaddrtxt);
        }

```

tboegi (changes requested):

asyn/drvAsynSerial/drvAsynIPServerPort.c
lrossa: Yes, it is: for an IPv4 address, 23 (or 32) bytes are sufficient. For an IPv6 address you need at least 49 bytes ("[aaaa:bbbb:cccc:dddd:eeee:ffff:gggg:hhhh]:ppppp" plus NUL character), so I used a buffer size of 50.

documentation/asynDriver.html
lrossa: So should it test for a valid IP address and use it? Otherwise a name or invalid IP address would behave like before...

asyn/drvAsynSerial/drvAsynIPServerPort.c
lrossa: From my point of view, it should be possible to bind only to the loopback interface. A special name as exception could produce confusion.

asyn/drvAsynSerial/drvAsynIPServerPort.c
tboegi: Did you reserve some bytes for the interfaceID ?
The "%eno1" from below:
[fe80::2547:708e:a277:b32c%eno1]

documentation/asynDriver.html
tboegi: May be I wasn't clear enough:
What happens if someone specifies
nonexistinghost:4002
I would expect asyn to fail loudly instead of silently falling back to INADDR_ANY.
This is debatable.

But at minimum I would like to see a warning in the IOC log

lrossa:
Now the previous examples should work as expected. If you provide an invalid name or address, the call should complain and fail. I increased the buffer size, which might include an interfaceID. What is your option?

lrossa:
_changed description to:_
The 2nd parameter of "drvAsynIPServerPortConfigure" allows to specify a
host name or ip address for multi-homed machines. If you want to specify
any interface, use the example "localhost:port", an empty host ":port" or
the any address "0.0.0.0:port". The address has to be an existing interface
on the host. If the name cannot be resolved, the call will complain and
fail. If you need the loopback interface, use "127.0.0.1:port".

PR #156, opened 2022-05-18
Add additional modules

FreddieAkeroyd:
Add additional modules for CI, update module versions to base 7.0 compatible ones

### DIFF
--- a/.ci-local/base3-14.set
+++ b/.ci-local/base3-14.set
@@ -1,4 +1,7 @@
-MODULES="sncseq"
+MODULES="sncseq calc sscan ipac"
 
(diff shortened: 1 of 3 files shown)


FreddieAkeroyd:
builds pass except on 32bit mingw as epics base fails to build due to `epicsTime.obj:epicsTime.cpp:(.text+0x9f2): undefined reference to _mkgmtime32`

FreddieAkeroyd:
I think this passes on epics-base build, maybe the `.ci` directory needs updating as Appveyor did have a choice of mingw installs and this may have changed in a later `.ci`

PR #158, opened 2022-07-08
Fix valueChanged flag for UInt32 paramVal.

ericonr:
The other paramVal types set the valueChanged flag if the parameter goes
from undefined to defined, but setUInt32 doesn't. Therefore, we need to
add a conditional to do this. Since the value is starting off as
undefined, we make sure to set data.uival to 0, since it's going to be
used to generate the callback mask as well.

This bug was detected when using setUIntDigitalParam() from the
asynPortDriver class, where setting a parameter to 0 still leaves the
corresponding PV undefined (per camonitor output), because the callback
wasn't called.

Since we are here, also:

- fix indentation to fit rest of file (4 spaces instead of 2)
- simplify application of mask: simply zero the bits which the mask has
  set, then OR the result with the masked value

### DIFF
--- a/asyn/asynPortDriver/paramVal.cpp
+++ b/asyn/asynPortDriver/paramVal.cpp
@@ -189,24 +189,38 @@ epicsInt64 paramVal::getInteger64()
   * \return Returns asynParamBadIndex if the index is not valid or asynParamWrongType if the parameter type is not asynParamUInt32Digital. */
 void paramVal::setUInt32(epicsUInt32 value, epicsUInt32 valueMask, epicsUInt32 interruptMask)
 {
-    epicsUInt32 oldValue;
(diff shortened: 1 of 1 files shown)


MarkRivers:
> Since the value is starting off as undefined, we make sure to set data.uival to 0, since it's going to be used to generate the callback mask as well.

That should not really be necessary because the paramVal constructor has set the memory for the data union to 0.
```
paramVal::paramVal(const char *name, asynParamType type):
    type(type), status_(asynSuccess), alarmStatus_(0), alarmSeverity_(0),
    valueDefined(false), valueChanged(false){
    this->name = epicsStrDup(name);
    memset(&this->data, 0, sizeof(this->data));
}
```

But it is OK to set it to 0 again so I will merge this.

PR #159, opened 2022-08-25
Add IOC shell help text for asynSetTrace*Mask

mdavidsaver:
Recent Base allows IOC shell commands to provide a usage message.  This is signaled with the `IOCSHFUNCDEF_HAS_USAGE` processor macro to maintain backwards compatibility.

Add messages for `asynSetTraceMask` and `asynSetTraceIOMask` with the bit mask definitions (which I never remember).

```
epics> help asynSetTraceMask 

Usage: asynSetTraceMask portName addr mask

Set logging level bit mask.
  mask bits
    0x01 - Errors
    0x02 - Device I/O
    0x04 - Filter I/O
    0x08 - Driver I/O
    0x10 - Code Flow
    0x20 - Warnings
```

### DIFF
--- a/asyn/miscellaneous/asynShellCommands.c
+++ b/asyn/miscellaneous/asynShellCommands.c
@@ -621,7 +621,32 @@ static const iocshArg asynSetTraceMaskArg2 = {"mask", iocshArgString};
 static const iocshArg *const asynSetTraceMaskArgs[] = {
     &asynSetTraceMaskArg0,&asynSetTraceMaskArg1,&asynSetTraceMaskArg2};
 static const iocshFuncDef asynSetTraceMaskDef =
-    {"asynSetTraceMask", 3, asynSetTraceMaskArgs};
+    {"asynSetTraceMask", 3, asynSetTraceMaskArgs
+#ifdef IOCSHFUNCDEF_HAS_USAGE
+     ,
+#else
+    };
+static const char asynSetTraceMaskUsage[] =
+#endif
+     "Set logging level bit mask, either as a number,"
+     "  or since asyn R4-35 a combination of names.\n"
+     "  mask bits\n"
+     "    0x01 - ERROR\n"
+     "    0x02 - DEVICE\n"
+     "    0x04 - FILTER\n"
+     "    0x08 - DRIVER\n"
+     "    0x10 - FLOW\n"
+     "    0x20 - WARNING\n"
+     "\n"
+     "  eg. the following are equivalent\n"
+     "    asynSetTraceMask MYPORT -1 0x21\n"
+     "    asynSetTraceMask MYPORT -1 \"WARNING+ERROR\"\n"
+#ifdef IOCSHFUNCDEF_HAS_USAGE
+    };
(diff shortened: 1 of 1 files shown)


MarkRivers:
Starting with asyn R4-35 in March, 2019 it is no longer necessary to use the bit mask values.  You can do this:
```
asynSetTraceMask port addr ERROR|DRIVER
asynSetTracceIOMask port addr HEX|ESCAPE
```
I would suggest using these symbols in the Usage output.  They are much easier to remember than the bit mask values.

mdavidsaver:
How about the following?  (which @MarkRivers can edit)

For asynSetTraceMask
```
Set logging level bit mask, either as a number, or since R4-35 a combination of names.
  mask bits
    0x01 - ERROR
    0x02 - DEVICE
    0x04 - FILTER
    0x08 - DRIVER
    0x10 - FLOW
    0x20 - WARNING

  eg. the following are equivalent
    asynSetTraceMask MYPORT -1 0x21
    asynSetTraceMask MYPORT -1 "WARNING|ERROR"
```
For asynSetTraceIOMask
```
Set logging I/O printing formatting bit mask, either as a number, or since R4-35 a combination of names.
  mask bits
    0x01 - ASCII
    0x02 - ESCAPE
    0x04 - HEX

  eg. the following are equivalent
    asynSetTraceIOMask MYPORT -1 0x2
    asynSetTraceIOMask MYPORT -1 "ESCAPE"
```

mdavidsaver:
Updated.

MarkRivers:
Since this message will only appear on systems running asyn R4-43 and later it seems like the phrase about R4-35 is not needed.

MarkRivers:
AppVeyor is failing with this error:
```
devAsynOctet.c
C:\projects\asyn\asyn\devEpics\devAsynOctet.c(57): fatal error C1083: Cannot open include file: 'sCalcoutRecord.h': No such file or directory
make[2]: *** [C:/Users/appveyor/.cache/base-7.0/configure/RULES_BUILD:259: devAsynOctet.obj] Error 2
make[2]: Leaving directory 'C:/projects/asyn/asyn/O.windows-x64'
make[2]: *** Waiting for unfinished jobs....
make[2]: Entering directory 'C:/projects/asyn/asyn/O.windows-x64'
```
But that include is protected:
```
#ifdef HAVE_CALCMOD
#include "sCalcoutRecord.h"
#endif /* HAVE_CALCMOD */
```
HAVE_CALCMOD should only be defined if CALC is defined in configure/RELEASE.

Why is this error happening?

mdavidsaver:
> Since this message will only appear on systems running asyn R4-43 and later it seems like the phrase about R4-35 is not needed.

"needed" no.  I do think it is worth mention as it might help someone understand a "works for me" situation a little more quickly.

wrt. documentation, I'm coming to favor an approach of trying to put more information closer to where a user will see it during "normal" usage.

As I think about it.  By this logic, it would be handy to print this extended message as well for a parse error or out of range mask.

mdavidsaver:
Updated.  The usage message will also be printed before the error message when an invalid symbolic bit mask is given.

Issue #139, opened 2021-09-07
asynXXXArray should support aai and aao record types


ralphlange:
asyn only supports the waveform record for array data.

The aai and aao records have been part of EPICS Base for a long time. They provide proper input and output functionality for array data and are similar enough to the waveform record to be supported without much effort.

vnadot:
Hi @ralphlange ,

I'm indeed looking for a way to write an array with asyn (for my FPGA). What is your trick without this asyn feature ?

The array read works fine for me:
```
record(waveform, "$(P=CEA:)FileReader0DataRb") {
    field(DESC, "File reader 0 data")
    field(DTYP, "asynFloat64ArrayIn")
    field(FTVL, "DOUBLE")
    field(NELM, "${NELM=4096}")
    field(SCAN, "I/O Intr")
    field(INP,  "@asyn(${PORT})FileReader0Data")
}
```

ralphlange:
Asyn also supports `asyn<TYPE>ArrayOut` for the waveform record.
As simple as that.

vnadot:
Indeed, it works fine like this, thanks !

```
record(waveform, "${P=CEA:}FileReader0DataSet") {
    field(DESC, "File reader 0 data")
    field(DTYP, "asynFloat64ArrayOut")
    field(FTVL, "DOUBLE")
    field(NELM, "${NELM=4096}")
    field(INP,  "@asyn(${PORT})FileReader0Data")
    field(EGU,  "${EGU=V}") # should be set by saverestoreApp ?
    field(PREC, "${PREC=3}")
}
record(waveform, "${P=CEA:}FileReader0DataRb") {
    field(DESC, "File reader 0 data")
    field(DTYP, "asynFloat64ArrayIn")
    field(FTVL, "DOUBLE")
    field(NELM, "${NELM=4096}")
    field(SCAN, "I/O Intr")
    field(INP,  "@asyn(${PORT})FileReader0Data")
    field(EGU,  "${EGU=V}") # should be set by saverestoreApp ?
    field(PREC, "${PREC=3}")
}
```

ralphlange:
Sure.
It's more an annoyance that EPICS Base has proper array-in and array-out records, but asyn ignores them. Under the hood, it's all there.

MarkRivers:
I would like to add support for aao and aai.  But I want to do it without needing to write a large amount new code that is repetitive with the waveform record support.

[...]

PR #162, opened 2022-09-17
Improve waveform device support, add aai and aao

MarkRivers:
This PR addresses #139.

Previously devAsynXXXArray used a 485 line C macro.  This has been rewritten to use a new C++ template class in devAsynXXXArray.cpp. The new class adds support for aao and aai records, in addition to the existing waveform record support.  The device support creates 24 dsets (6 data types * 4 record types).

Data types:
- int8
- int16
- int32
- int64 (where supported)
- float32
- float64

Record types:
- waveform in
- waveform out
- aai
- aao

The new code has 9 fewer files and 150 fewer lines of code than R4-43.

It still has a 30 line C macro which is used to create the 120 C functions used by the 24 dsets.  Any ideas on whether that could also be made more elegant would be welcome.


asyn/devEpics/devAsynXXXArray.cpp
```diff
@@ -0,0 +1,573 @@
...
+        (devAsynXXXArray<REC, INTERFACE, INTERRUPT, EPICS_TYPE> *) drvPvt;                 \
+    pObj->interruptCallback(pasynUser, value, len);                                        \
+}                                                                                          \
+static long INIT_FUNC(dbCommon *pr) {                                                      \
+    new devAsynXXXArray<REC, INTERFACE, INTERRUPT, EPICS_TYPE>                             \
+                        (pr, &((REC *)pr)->LINK, SIGNED_TYPE, UNSIGNED_TYPE, IS_OUTPUT,    \
+                        INTERFACE_NAME, QRCB_FUNC, INTCB_FUNC);                            \
+    return 0;                                                                              \
+}                                                                                          \
+analogDset DSET = {6, 0, 0, INIT_FUNC, GET_INFO_FUNC, PROC_FUNC, 0};                       \
```
mdavidsaver, line 484: The `analogDset DSET = ...` definition should be `static`.

asyn/devEpics/devAsynXXXArray.cpp
```diff
@@ -0,0 +1,573 @@
...
+epicsExportAddress(dset, DSET);
+
+// 8-bit integer arrays
+MAKE_DEVSUP(asynInt8ArrayWfIn, waveformRecord, inp, asynInt8Array, "asynInt8Array", interruptCallbackInt8Array, epicsInt8,
```
mdavidsaver, line 488: With some target configuration, `epicsExportAddress()` must be wrapped in `extern "C"`.  The simplest way to do this would be to enclose all of the `MAKE_DEVSUP()` expansions.  eg. `extern "C" { MAKE_DEVSUP(...); ... }`.

asyn/devEpics/devAsynXXXArray.cpp
```diff
@@ -0,0 +1,573 @@
...
+                else
+                    scanIoRequest(ioScanPvt_);
+            }
+            epicsMutexUnlock(ringBufferLock_);
+        }
+    }
+};
+
+extern "C" {
+typedef struct analogDset { /* analog  dset */
```
mdavidsaver, line 445: I don't think this `extern "C"` should start so early.

Also, with C++ the `typedef struct x {...} x;` pattern is not necessary.  A plain `struct x {...};` is equivalent.

For that matter, I think that the definitions of `struct analogDset` and `class devAsynXXXArray` should be placed in an [anonymous namespace](https://en.cppreference.com/w/cpp/language/namespace#Unnamed_namespaces) to remove any possibility of a name collision (although I think this is already unlikely).

An anonymous namespace (aka. `namespace { ... }`) is the c++ super-set of a C `static` global, which also [hides](https://en.cppreference.com/w/cpp/language/storage_duration#internal_linkage) some symbols which are defined implicitly.  eg. type and exception unwinding info.

eg. in `namespace { static int myglobal; }` the `static` qualifier is redundant.

One relevant example, where [all but](https://github.com/mdavidsaver/pvxs/blob/f0de452930235ae58ee0bf059a816a222b82a9fc/ioc/iochooks.cpp#L25-L243) the [last two](https://github.com/mdavidsaver/pvxs/blob/f0de452930235ae58ee0bf059a816a222b82a9fc/ioc/iochooks.cpp#L245-L261) symbols are hidden.

The overall structure of `devAsynXXXArray.cpp` would then be

```c++
// #include s
namespace {
template<...> class devAsynXXXArray { ... };
struct analogDset { ... };
} // end of namespace
#define MAKE_DEVSUP(...) ...
extern "C" {
MAKE_DEVSUP(...); ...
/// MAKE_DEVSUP expands to only static globals or epicsExport*()
} // end of extern "C"
```

With this structure, running `nm --defined-only --extern-only devAsynXXXArray.o` should list only `pvar_*` symbols as strong definitions (`D`).  The internal symbols will either be (`d` or `t`), or optimized away entirely.  (`man nm` lists all of the letter codes)

mdavidsaver (commented):
Neat.  I have a couple of comments on the overall structure of the new `devAsynXXXArray.cpp` file.

mdavidsaver:
[Here](https://github.com/mdavidsaver/pva2pva-old/blob/4559f483bf092f3cd1086f4758ceb1015a9d288c/pdbApp/demo.cpp#L69-L71) is another example of `namespace {} extern "C" {}` which may be easier to parse.

asyn/devEpics/devAsynXXXArray.cpp
ralphlange: +1
That's exactly how I am structuring/declaring C++ device support nowadays.

ralphlange (commented):
Don't have anything to add to Michael's comments. (Thanks!)

anjohnson (approved):
This looks like it should simplify maintenance in the future.

The only comment I have is that the new devAsynXXXArray.cpp file could use a copyright / license notice at the top, it's better to make the licensing terms explicit in the more complex source files at least.

MarkRivers:
Thanks all for the review, I have made the changes @mdavidsaver and @anjohnson suggested.

PR #168, opened 2023-01-15
GHA update

mdavidsaver:
GHA builds on `ubuntu-latest` are currently failing for lack of `TIRPC=YES`.  Also, update some actions to avoid deprecation warnings.

Is it time to drop the 3.14 job?

The `BASE=3.14` job continues to fail while building epics-base with an error which I don't understand.  I am able to successfully build the 3.14 branch locally.  Recent [advice on tech-talk](https://epics.anl.gov/tech-talk/2022/msg01561.php) is that `The 3.14 release series is dead ...`.

```
...
  In file included from ../../../../src/cas/generic/caServer.cc:21:
  ../../../../src/cas/generic/caServerI.h:92:5: error: ‘class tsDLList<ioBlocked> tsDLList<ioBlocked>::tsDLList’ is private within this context
     92 |     tsDLList < casIntfOS > intfList;
        |     ^~~~~~~~~~~~~~~~~~~~~~
  In file included from ../../../../src/cas/generic/caServerI.h:33,
                   from ../../../../src/cas/generic/caServer.cc:21:
  ../../../../src/cas/generic/ioBlocked.h:43:7: note: declared private here
     43 | class ioBlockedList : private tsDLList<ioBlocked> {
        |       ^~~~~~~~~~~~~
```

PR #169, opened 2023-01-15
asynDbGetInfo: consolidate dbGetInfo and plug leak at init

mdavidsaver:
Avoid leaking `DBENTRY` during initialization.

Prefer constant time `dbInitEntryFromRecord()` to `dbFindRecord()` when available.

Introduce devEpicsPvt.h/.c to avoid some duplication


mdavidsaver:
The GHA builds are failing pending #168.  The `testErrors` IOC starts, and I've used GDB to verify that `asynDbGetInfo()` returns `!NULL` when an info take is present (against Base 7.0).

MarkRivers:
I am assuming this will pass the tests now that I have merged the change for TIRPC, so I am merging it.

PR #178, opened 2023-04-28
Misc fixes found by static code analysis

ralphlange:
A set of fixes for different issues (few major, many minor) that were found by cppcheck/sonar at ITER.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ O.*/
 *.swp
 *_BAK.adl
 /QtC-*
+/.qtc_*
 cdCommands
 envPaths
 dllPath.bat
--- a/asyn/asynDriver/asynManager.c
+++ b/asyn/asynDriver/asynManager.c
@@ -3000,7 +3000,7 @@ static int traceVprintIOSource(asynUser *pasynUser,int reason,
         }
     }
     /* If the traceIOMask is 0 or traceTruncateSize <=0 we need to output a newline */
-    if((traceIOMask == 0) || (traceTruncateSize <=0)) {
+    if((traceIOMask == 0) || (traceTruncateSize ==0)) {
         if(fp) {
             nout += fprintf(fp,"\n");
         } else {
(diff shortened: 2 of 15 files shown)


MarkRivers (commented):
@ralphlange this looks good thanks.  

My only comment is that changing %u to %lu for size_t arguments is still not correct on 64-bit Windows, right?  There a size_t is 64-bit but a long is 32-bit.  Changing is to *llu won't be correct on 32-bit machines since a size_t is 32-bit and a long long is 64-bit?  Should we cast the argument instead?

norumwe12:
I presume this is an issue because there are platforms that don’t support the ‘z’ modifier for size_t  (“%zu”) nor the “PRI” string literals?
Perhaps some EPICS OS-dependent header could provide the PRIxxx literals for such machines?


> On Apr 29, 2023, at 8:06 AM, Mark Rivers ***@***.***> wrote:
> 
> My only comment is that changing %u to %lu for size_t arguments is still not correct on 64-bit Windows, right? There a size_t is 64-bit but a long is 32-bit. Changing is to *llu won't be correct on 32-bit machines since a size_t is 32-bit and a long long is 64-bit? Should we cast the argument instead?
> 
> 

— 
Eric Norum
***@***.***

MarkRivers:
> I presume this is an issue because there are platforms that don’t support the ‘z’ modifier for size_t  (“%zu”) nor the “PRI” string literals?
Perhaps some EPICS OS-dependent header could provide the PRIxxx literals for such machines?

This has been discussed in this tech-talk thread.  Apparently of the supported architectures only vxWorks does not support %z.
https://epics.anl.gov/core-talk/2022/msg00321.php

ralphlange:
True, I didn't consider that.
Guess the only waterproof option is to cast the argument to the widest in-the-wild type that we need to support. Anything else would either fail or be insufficient on some architectures. Only if that widest type is not supported by all compilers, we'd be in trouble.

MarkRivers:
I thing cast to (long long) and %llu would work on all?

ralphlange:
Worth a try.
Better than %z that would fail hard on VxWorks, while the other mismatches would just generate warnings.

PR #180, opened 2023-05-02
Add support for sending serial break via option interface.

lrossa:
The asyn option interface is used to set/clear a break state on serial lines.

Thanks to idea how to do it under Microsoft Windows are taken from here:
  Freddie Akeroyd <freddie.akeroyd@stfc.ac.uk>
  https://github.com/ISISComputingGroup/EPICS-asyn
  commit 1228f8ea40fcc1060fd996b4a8669c3572d1d4ba (Tue Nov 14 12:57:31 2017 +0000)
  commit a8496cf53a4b138469a639cb7bbc59caea8c3ec6 (Tue Nov 14 13:08:37 2017 +0000)


FreddieAkeroyd:
Hi @Irossa in https://github.com/ISISComputingGroup/EPICS-asyn/pull/5 a serial break was applied for a fixed duration automatically at the end of a `writeIt` as we had a device that used this as a sort of command terminator. If I understand correctly, in your case you only need to apply/clear a serial break via a `setOption` command? I'd like to submit a PR to add this automatic break functionality too, I wasn't sure of the best naming scheme though. I could use `autobreak` and "yes/no/number" like your `break` command, or add a `break_len` option and `break` then takes "on/off/auto". Thank you for adding RFC2217 support in asynInterposeCom.c - we don't use that protocol at the moment, but it is something we have been looking at.

FreddieAkeroyd:
Does something along the lines of #188 look reasonable?

PR #182, opened 2023-06-02
allow 3 EOS characters rather than 2

rerpha:
Hi, we weren't sure if there was a reason why asyn only seemed to allow 2 chars for the EOS/terminator string. We spotted a comment on line 217, but can't quite figure out what it's trying to do - are there any other areas which expect or rely on the EOS chars being 2 characters?

### DIFF
--- a/asyn/miscellaneous/asynInterposeEos.c
+++ b/asyn/miscellaneous/asynInterposeEos.c
@@ -44,13 +44,13 @@ typedef struct eosPvt {
     char          *inBuf;
     unsigned int  inBufHead;
     unsigned int  inBufTail;
-    char          eosIn[2];
+    char          eosIn[3];
     int           eosInLen;
     int           eosInMatch;
     int           processEosOut;
     size_t        outBufSize;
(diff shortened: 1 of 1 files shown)


MarkRivers:
The comment in line 217 is the key to the problem.  The asynOctet drivers get data from the device in "chunks".  The problem arises when a potential terminator begins on the last character of the current chunk, and might continue on the first character of the next chunk.  The logic currently implemented correctly handles the case of a 2-character terminator.  The logic gets more complicated if the terminator can be longer than 2 characters, since we need to keep track of how many characters of the potential terminator have been received.  It gets messy if the next chunk is only 1 character.   When the EOS limit is 2 characters it is pretty easy, since the next chunk will contain at least 1 character.  But if the EOS limit is more than 2 then we need to look at the following chunks as well. Real serial port drivers often return data 1 character at a time, so this is not just a hypothetical problem.

I agree it would be nice to do this, and it is not impossible, it just requires careful thought and testing.

PR #190, opened 2023-07-20
Allow make build to run correctly

### LINKED ISSUE #189: `make build` fails
If you do the following:
* Install EPICS base (7.0 branch)
* Try to run `make build` in asyn after configuring it to point at EPICS base

then you will have a failed build failing with
```
make[1]: *** No rule to make target `devAsynOctet.dbd', needed by `/devEpics.dbd'.  Stop.
make: *** [asyn.build] Error 2
```
The reason appears to be the following code from asyn/Makefile (https://github.com/epics-modules/asyn/blob/master/asyn/Makefile#L340):
```makefile
ifdef DBDCAT_COMMAND
  # Override the default command when generating devEpics.dbd
  $(COMMON_DIR)/devEpics.dbd: DBDCAT_COMMAND = \
      $(PERL) $(TOOLS)/makeIncludeDbd.pl $(devEpics_DBD) $(notdir $@)
else
  # 3.14 didn't have the DBDCAT rule:
  build: $(COMMON_DIR)/devEpics.dbd
  $(COMMON_DIR)/devEpics.dbd: $(devEpics_DBD)
	$(ECHO) "Creating dbd file $(notdir $@)"
	@$(RM) $(notdir $@)
	$(PERL) $(TOOLS)/makeIncludeDbd.pl $(devEpics_DBD) $(notdir $@)
	@$(MV) $(notdir $@) $@
endif
```
From what I can see, the issue is that on the first pass through in EPICS base (https://github.com/epics-base/epics-base/blob/7.0/configure/RULES) `T_A` is not defined which means we do not define `DBDCAT_COMMAND` (from `RULES.Db`), which adds the dependency
```makefile
build: $(COMMON_DIR)/devEpics.dbd
```
However, on this pass `COMMON_DIR` is _also_ not yet defined, so we are saying that `build` depends on `/devEpics.dbd`, which of course there is no rule to create.

A few comments:
* I'm not sure this is an asyn issue per se; this is definitely about the interaction between asyn's (custom) build rules and EPICS base.
* The build instructions for `asyn` explicitly say that you should run `make` and not `make build`, so _technically_ I am not following the instructions. However, for EPICS base `make` and `make build` do the same thing; moreover, `make build` is not an _unreasonable_ command to try to run, and it is surprising that it fails while `make` or even `make install` succeed.

simon-ess:
Closes #189 

This simply encloses the custom build rules into the `T_A`-dependent portion; these rules should only be relevant once the architecture is defined, so this should be the simplest fix.

### DIFF
--- a/asyn/Makefile
+++ b/asyn/Makefile
@@ -305,7 +305,7 @@ drvVxi11$(DEP): vxi11intr.h vxi11core.h
 vxi11core_xdr.c$(DEP): vxi11core.h
 
 ifdef T_A
-  ifneq ($(findstring $(OS_CLASS),vxWorks RTEMS),)
+ ifneq ($(findstring $(OS_CLASS),vxWorks RTEMS),)
 
 vxi11core.h vxi11core_xdr.c vxi11intr.h vxi11intr_xdr.c : \
 % : ../vxi11/rpc/%
@@ -313,7 +313,7 @@ vxi11core.h vxi11core_xdr.c vxi11intr.h vxi11intr_xdr.c : \
 vxi11core_xdr.o: vxi11core.h
 vxi11intr_xrd.o: vxi11intr.h
 
-  else
+ else
 
 .SECONDARY: vxi11core_xdr.c vxi11intr_xdr.c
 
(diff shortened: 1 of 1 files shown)


simon-ess:
Note that the builds failing seems to be largely due to the unavailability of sequencer at the moment. 😢

PR #200, opened 2023-12-20
Improve connection cleanup to avoid many error messages with ioc exit on some operating systems

FreddieAkeroyd:
Try to terminate accept thread during cleanup

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPServerPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPServerPort.c
@@ -305,7 +305,7 @@ static void connectionListener(void *drvPvt)
     asynPrint(pasynUser, ASYN_TRACE_FLOW,
             "drvAsynIPServerPort: %s started listening for connections on %s\n",
             tty->portName, tty->serverInfo);
-    while (1) {
+    while (tty->fd != INVALID_SOCKET) {
         if (tty->socketType == SOCK_DGRAM) {
             if ((tty->UDPbufferPos == 0) && (tty->UDPbufferSize == 0)) {
                 tty->UDPbufferSize = recvfrom(tty->fd, tty->UDPbuffer, THEORETICAL_UDP_MAX_SIZE , 0, NULL, NULL);
@@ -498,6 +498,7 @@ static void ttyCleanup(void *pPvt)
     if (tty->fd >= 0) {
         asynPrint(tty->pasynUser, ASYN_TRACE_FLOW, "drvAsynIPServerPort:ttyCleanup %s: shutdown socket %d\n", tty->portName, tty->fd);
         epicsSocketDestroy(tty->fd);
+        tty->fd = INVALID_SOCKET;
     }
     free(tty->portName);
     free(tty->serverInfo);
--- a/asyn/drvAsynSerial/drvAsynIPServerPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPServerPort.c
@@ -499,6 +499,7 @@ static void ttyCleanup(void *pPvt)
         asynPrint(tty->pasynUser, ASYN_TRACE_FLOW, "drvAsynIPServerPort:ttyCleanup %s: shutdown socket %d\n", tty->portName, tty->fd);
         epicsSocketDestroy(tty->fd);
         tty->fd = INVALID_SOCKET;
+        epicsThreadSleep(0.1); /* wait for accept thread to terminate */
     }
     free(tty->portName);
     free(tty->serverInfo);
--- a/asyn/drvAsynSerial/drvAsynIPServerPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPServerPort.c
@@ -324,6 +324,12 @@ static void connectionListener(void *drvPvt)
 
         } else {
             clientFd = epicsSocketAccept(tty->fd, (struct sockaddr *) &clientAddr, &clientLen);
+            if (tty->fd == INVALID_SOCKET) {
+                asynPrint(pasynUser, ASYN_TRACE_FLOW,
+                    "drvAsynIPServerPort: terminating connection thread for %s\n",
+                    tty->serverInfo);                
+                break; /* we must be in ioc shutdown and ttyCleanup has been called */
+            }
             asynPrint(pasynUser, ASYN_TRACE_FLOW,
                     "drvAsynIPServerPort: new connection, socket=%d on %s\n",
                     clientFd, tty->serverInfo);
--- a/asyn/drvAsynSerial/drvAsynIPServerPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPServerPort.c
@@ -295,8 +295,6 @@ static void connectionListener(void *drvPvt)
     int i;
     portList_t *pl, *p;
     int connected;
-    fd_set read_set;
-    struct timeval select_timeout;
 
     /*
      * Sanity check
@@ -308,13 +306,6 @@ static void connectionListener(void *drvPvt)
             "drvAsynIPServerPort: %s started listening for connections on %s\n",
             tty->portName, tty->serverInfo);
     while (tty->fd != INVALID_SOCKET) {
-        FD_ZERO(&read_set);
-        FD_SET(tty->fd, &read_set);
-        select_timeout.tv_sec  = 0;
-        select_timeout.tv_usec = 50000; /* 0.05 seconds, needs to be less than delay in ttyCleanup */
-        if (select(tty->fd + 1, &read_set, NULL, NULL, &select_timeout) < 1 || !FD_ISSET(tty->fd, &read_set)) {
-            continue;
-        }
         if (tty->socketType == SOCK_DGRAM) {
             if ((tty->UDPbufferPos == 0) && (tty->UDPbufferSize == 0)) {
                 tty->UDPbufferSize = recvfrom(tty->fd, tty->UDPbuffer, THEORETICAL_UDP_MAX_SIZE , 0, NULL, NULL);
@@ -506,9 +497,13 @@ static void ttyCleanup(void *pPvt)
     if (!tty) return;
     if (tty->fd >= 0) {
         asynPrint(tty->pasynUser, ASYN_TRACE_FLOW, "drvAsynIPServerPort:ttyCleanup %s: shutdown socket %d\n", tty->portName, tty->fd);
-        epicsSocketDestroy(tty->fd);
+        SOCKET fd_save = tty->fd;
         tty->fd = INVALID_SOCKET;
-        epicsThreadSleep(0.1); /* wait for accept thread to terminate */
+        /* shutdown then close the socket so accept thread will terminate */
+        shutdown(fd_save, SHUT_RDWR);
+        epicsThreadSleep(0.01); /* give time for shutdown to be noticed prior to close */
+        epicsSocketDestroy(fd_save);
+        epicsThreadSleep(0.1);
     }
     free(tty->portName);
     free(tty->serverInfo);
--- a/asyn/drvAsynSerial/drvAsynIPServerPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPServerPort.c
@@ -295,6 +295,8 @@ static void connectionListener(void *drvPvt)
     int i;
     portList_t *pl, *p;
     int connected;
+    fd_set read_set;
+    struct timeval select_timeout;
 
     /*
      * Sanity check
@@ -306,6 +308,13 @@ static void connectionListener(void *drvPvt)
             "drvAsynIPServerPort: %s started listening for connections on %s\n",
             tty->portName, tty->serverInfo);
     while (tty->fd != INVALID_SOCKET) {
+        FD_ZERO(&read_set);
+        FD_SET(tty->fd, &read_set);
+        select_timeout.tv_sec  = 0;
+        select_timeout.tv_usec = 50000; /* 0.05 seconds, needs to be less than delay in ttyCleanup */
+        if (select(tty->fd + 1, &read_set, NULL, NULL, &select_timeout) < 1 || !FD_ISSET(tty->fd, &read_set)) {
+            continue;
+        }
         if (tty->socketType == SOCK_DGRAM) {
             if ((tty->UDPbufferPos == 0) && (tty->UDPbufferSize == 0)) {
                 tty->UDPbufferSize = recvfrom(tty->fd, tty->UDPbuffer, THEORETICAL_UDP_MAX_SIZE , 0, NULL, NULL);


exzombie:
I like the general idea. It won't work on Linux, however. It might on Darwin, or it might not; I don't have a Mac to test on. The reason is that `epicsSocketDestroy()` on Posix only calls `close()` on the socket, which is not always enough to interrupt the blocking `accept()`. One needs to call `shutdown()` before calling `close()`.[^shutdestroy]

[^shutdestroy]: Perhaps it would be appropriate to add `shutdown()` to `epicsSocketDestroy()`? Not sure.

@jlmuir, if your're testing this, and if it works, you should expect to see one or two errors that occur when the `accept()` is interrupted. If it doesn't work, add `shutdown(tty->fd, SHUT_RDWR);` before destroying the socket. Once you've confirmed that this helps on Darwin, we can polish it to remove the remaining errors.

FreddieAkeroyd:
Ah thanks, I see why linux did not error then, it just stayed blocked until the process died.  Maybe adding `shutdown` to `epicsSocketDestroy` is worth doing, might be work a post to `core-talk` to see if it was considered but decided against. Can `shutdown` itself ever block though? 

Another option is to make the loop non-blocking by adding a select() with timeout, so accept is only called if there is a connection to accept, and fd can then be checked every timeout

jlmuir:
@exzombie, I confirm that the three commits from @FreddieAkeroyd in this PR help on `darwin-aarch64`!  I now only get one error message on IOC exit as you predicted:

```
2023/12/21 12:21:18.686 drvAsynIPServerPort:ttyCleanup P5001: shutdown socket 4
```

Here's the full terminal session:
```
$ ../../bin/darwin-aarch64/testIPServer st.cmd
< envPaths
epicsEnvSet("IOC","ioctestIPServer")
epicsEnvSet("TOP","/tmp/48/opt/synapps-6.2.1/asyn-R4-44-2")
epicsEnvSet("SUPPORT","/tmp/48/opt/synapps-6.2.1")
epicsEnvSet("SNCSEQ","/tmp/48/opt/synapps-6.2.1/seq-2-2-9")
epicsEnvSet("CALC","/tmp/48/opt/synapps-6.2.1/calc-R3-7-4")
epicsEnvSet("SSCAN","/tmp/48/opt/synapps-6.2.1/sscan-R2-11-4")
epicsEnvSet("EPICS_BASE","/tmp/48/opt/epics-base-7.0.8")
dbLoadDatabase("../../dbd/testIPServer.dbd")
testIPServer_registerRecordDeviceDriver(pdbbase)
#The following command starts a server on port 5001
drvAsynIPServerPortConfigure("P5001","localhost:5001",2,0,0,0)
serverAddr: 0.0.0.0:5001
serverPort: 5001
drvAsynIPServerPortConfigure("P5002","localhost:5002",1,0,0,0)
serverAddr: 0.0.0.0:5002
serverPort: 5002
#asynSetTraceFile("P5001",-1,"")
asynSetTraceIOMask("P5001",-1,0x2)
asynSetTraceMask("P5001",-1,0xff)
dbLoadRecords("../../db/testIPServer.db", "P=testIPServer:")
iocInit()
Starting iocInit
############################################################################
## EPICS R7.0.8
## Rev. 2023-12-18T15:15-0600
## Rev. Date build date/time:
############################################################################
iocRun: All initialization complete
ipEchoServer("P5001")
2023/12/21 12:20:57.598 P5001 -1 registerInterruptUser
seq("ipSNCServer", "P=testIPServer:, PORT=P5002")
sevr=info Sequencer release 2.2.9, compiled Mon Dec 18 22:30:46 2023
sevr=info Spawning sequencer program "ipSNCServer", thread 0x600002614180: "ipSNCServer"
sevr=info ipSNCServer[0]: all channels connected & received 1st monitor
epics> exit
2023/12/21 12:21:18.686 drvAsynIPServerPort:ttyCleanup P5001: shutdown socket 4
```

MarkRivers:
The appveyor Windows builds are failing because it cannot find "distutils".  Does anyone know how to fix this?

jlmuir:
Re the AppVeyor build failures, if the version of Python in the AppVeyor build VM is >= 3.12, maybe because the `distutils` module was removed from Python >= 3.12?  See [What’s New In Python 3.12](https://docs.python.org/3/whatsnew/3.12.html), which has more info on the removal, including a migration guide and workaround.

MarkRivers:
I found that the latest version of ci-scripts (v3.4.1) removes the dependency on distutils.  I have changed asyn to use the latest version of ci-scripts, and I suspect that will fix the problem with Appveyor.

FreddieAkeroyd:
@jlmuir good to hear that works, though on reflection I think the suggestion of @exzombie to use `shutdown()` of the socket prior to `epicsSocketDestroy` would be a better way to go as it avoids arbitrary wakeups of the accept loop and also allows the sleep delay in cleanup to be as small as possible. Something like this in the ttyCleanup and no select() in the accept loop
```
int fd_save = tty->fd;
tty->fd = INVALID_SOCKET;
shutdown(fd_save, SHUT_RDWR); 
epicsSocketDestroy(fd_save);
epicsThreadSleep(0.1); /* wait for accept thread to terminate */  
```

FreddieAkeroyd:
Though we may need a delay between `shutdown` and epicsSocketDestroy` if i read https://epics.anl.gov/core-talk/2019/msg00714.php correctly. I was also unaware of `epicsSocketSystemCallInterruptMechanismQuery`

FreddieAkeroyd:
Have pushed change to use `shutdown` prior to socket `close`, I believe this should mean that windows/linux/darwin will now all behave the same and print a single accept error on ioc exit?

jlmuir:
Tried with commit 7f89753 on `darwin-aarch64`, and it too eliminates the thousands of error messages (but on exit there are more messages than there were with the previous commit):

```
$ ../../bin/darwin-aarch64/testIPServer st.cmd
< envPaths
epicsEnvSet("IOC","ioctestIPServer")
epicsEnvSet("TOP","/tmp/48/opt/synapps-6.2.1/asyn-R4-44-2")
epicsEnvSet("SUPPORT","/tmp/48/opt/synapps-6.2.1")
epicsEnvSet("SNCSEQ","/tmp/48/opt/synapps-6.2.1/seq-2-2-9")
epicsEnvSet("CALC","/tmp/48/opt/synapps-6.2.1/calc-R3-7-4")
epicsEnvSet("SSCAN","/tmp/48/opt/synapps-6.2.1/sscan-R2-11-4")
epicsEnvSet("EPICS_BASE","/tmp/48/opt/epics-base-7.0.8")
dbLoadDatabase("../../dbd/testIPServer.dbd")
testIPServer_registerRecordDeviceDriver(pdbbase)
#The following command starts a server on port 5001
drvAsynIPServerPortConfigure("P5001","localhost:5001",2,0,0,0)
serverAddr: 0.0.0.0:5001
serverPort: 5001
drvAsynIPServerPortConfigure("P5002","localhost:5002",1,0,0,0)
serverAddr: 0.0.0.0:5002
serverPort: 5002
#asynSetTraceFile("P5001",-1,"")
asynSetTraceIOMask("P5001",-1,0x2)
asynSetTraceMask("P5001",-1,0xff)
dbLoadRecords("../../db/testIPServer.db", "P=testIPServer:")
iocInit()
Starting iocInit
############################################################################
## EPICS R7.0.8
## Rev. 2023-12-18T15:15-0600
## Rev. Date build date/time:
############################################################################
iocRun: All initialization complete
ipEchoServer("P5001")
2023/12/23 00:15:14.051 P5001 -1 registerInterruptUser
seq("ipSNCServer", "P=testIPServer:, PORT=P5002")
sevr=info Sequencer release 2.2.9, compiled Mon Dec 18 22:30:46 2023
sevr=info Spawning sequencer program "ipSNCServer", thread 0x600003bac0c0: "ipSNCServer"
sevr=info ipSNCServer[0]: all channels connected & received 1st monitor
epics> exit
2023/12/23 00:15:17.449 drvAsynIPServerPort: accept error on localhost:5002: fd=-1, Software caused connection abort
2023/12/23 00:15:17.554 drvAsynIPServerPort:ttyCleanup P5001: shutdown socket 4
2023/12/23 00:15:17.565 drvAsynIPServerPort: new connection, socket=-1 on localhost:5001
2023/12/23 00:15:17.565 drvAsynIPServerPort: accept error on localhost:5001: fd=-1, Undefined error: 0
```

FreddieAkeroyd:
Yes, it will print a few more messages as the `select` call avoided an error with `accept` and i didn't distinguish a select timeout (expected) from a socket close/shutdown. If we wished to avoid the `accept` error message on ioc exit we could use something like:
```
    clientFd = epicsSocketAccept(tty->fd, (struct sockaddr *) &clientAddr, &clientLen);
    if (tty->fd == INVALID_SOCKET) {
        break; /* we must be in ioc shutdown */
    }
```

jlmuir:
Tried with commit [19db7d3](https://github.com/epics-modules/asyn/pull/200/commits/19db7d33e5576393f99feeb196f88b7d321e5a95) on `darwin-aarch64`, and it too eliminates the thousands of error messages on exit, and now there are only two messages on exit:

```
$ ../../bin/darwin-aarch64/testIPServer st.cmd
< envPaths
epicsEnvSet("IOC","ioctestIPServer")
epicsEnvSet("TOP","/tmp/48/opt/synapps-6.2.1/asyn-R4-44-2")
epicsEnvSet("SUPPORT","/tmp/48/opt/synapps-6.2.1")
epicsEnvSet("SNCSEQ","/tmp/48/opt/synapps-6.2.1/seq-2-2-9")
epicsEnvSet("CALC","/tmp/48/opt/synapps-6.2.1/calc-R3-7-4")
epicsEnvSet("SSCAN","/tmp/48/opt/synapps-6.2.1/sscan-R2-11-4")
epicsEnvSet("EPICS_BASE","/tmp/48/opt/epics-base-7.0.8")
dbLoadDatabase("../../dbd/testIPServer.dbd")
testIPServer_registerRecordDeviceDriver(pdbbase)
#The following command starts a server on port 5001
drvAsynIPServerPortConfigure("P5001","localhost:5001",2,0,0,0)
serverAddr: 0.0.0.0:5001
serverPort: 5001
drvAsynIPServerPortConfigure("P5002","localhost:5002",1,0,0,0)
serverAddr: 0.0.0.0:5002
serverPort: 5002
#asynSetTraceFile("P5001",-1,"")
asynSetTraceIOMask("P5001",-1,0x2)
asynSetTraceMask("P5001",-1,0xff)
dbLoadRecords("../../db/testIPServer.db", "P=testIPServer:")
iocInit()
Starting iocInit
############################################################################
## EPICS R7.0.8
## Rev. 2023-12-18T15:15-0600
## Rev. Date build date/time:
############################################################################
iocRun: All initialization complete
ipEchoServer("P5001")
2023/12/28 22:46:47.218 P5001 -1 registerInterruptUser
seq("ipSNCServer", "P=testIPServer:, PORT=P5002")
sevr=info Sequencer release 2.2.9, compiled Mon Dec 18 22:30:46 2023
sevr=info Spawning sequencer program "ipSNCServer", thread 0x60000077c000: "ipSNCServer"
sevr=info ipSNCServer[0]: all channels connected & received 1st monitor
epics> exit
2023/12/28 22:46:53.865 drvAsynIPServerPort:ttyCleanup P5001: shutdown socket 4
2023/12/28 22:46:53.877 drvAsynIPServerPort: terminating connection thread for localhost:5001
```

FreddieAkeroyd:
@jlmuir thank you for testing and a happy new year to you. So now the only messages printed are from `ASYN_TRACE_FLOW`, so you will see them for port P5001 where this is enabled but not for P5002 where it is not (previously there was an `accept` error for P5002 that was printed, but such an error is now ignored during ioc shutdown)

exzombie:
Looks ok to me; I'm not a fan of using sleep to synchronize threads, but that's the least intrusive approach. I've had enough "fun" trying to fix broken (in theory) locking in #171 to push for a proper solution here 😅

I was thinking a bit about what it would mean if `shutdown()` was added to `epicsSocketDestroy()`. Would it break anything? I see three things to consider:
- The most visible change would be that threads on Linux would suddenly be unblocked on socket destruction, where they were blocked before. However, as we saw here, they get unblocked on some systems already, so the behavior would only change on Linux. This change would bring consistency, but might be a hard sell  because of the prevalence of Linux systems.
- The code that already calls `shutdown()` might be affected. The additional call made as part of `epicsSocketDestroy()` could cause errors. It does not on Linux (the second call returns success), but I don't know about other systems. Perhaps potential errors can simply be ignored, the shutdown being a "best effort" thing?
- There should be no delay between `shutdown()` and `close()`. From what I see, the RTEMS thing is either a bug/misbehavior or a thread priority problem or both. @mdavidsaver has added [a workaround](https://github.com/epics-base/pvAccessCPP/pull/169/files#diff-60215234c5f98e0b84d8b592fe7be9fa9e1d641968675b6e3b7a6867d518dd92R73) to pvAccessCPP, but I wonder whether it's really needed for other code. From the comments, it looks like the situation is specific to the design of pvAccessCPP.

Given the existence of the RTEMS workaround, surely adding `shutdown()` to `epicsSocketDestroy()` has been considered before? @mdavidsaver ?

mdavidsaver:
Since I am late to this party, these will have to be notes for future work.

> Given the existence of the RTEMS workaround, surely adding `shutdown()` to `epicsSocketDestroy()` has been considered before? @mdavidsaver ?

Not that I am aware of.

As was noted above, interrupting blocking syscalls is not especially portable to OS or socket type.  The awkward `epicsSocketSystemCallInterruptMechanismQuery()` logic is intended for deciding how to interrupt `recv()` on a TCP socket, but is I hope the correct condition for `accept()` as well.

asyn/drvAsynSerial/drvAsynIPServerPort.c
```diff
@@ -324,6 +324,12 @@ static void connectionListener(void *drvPvt)
 
         } else {
             clientFd = epicsSocketAccept(tty->fd, (struct sockaddr *) &clientAddr, &clientLen);
+            if (tty->fd == INVALID_SOCKET) {
+                asynPrint(pasynUser, ASYN_TRACE_FLOW,
+                    "drvAsynIPServerPort: terminating connection thread for %s\n",
+                    tty->serverInfo);                
+                break; /* we must be in ioc shutdown and ttyCleanup has been called */
```
mdavidsaver, line 331: ... or maybe the file descriptor limit was reached and accept() has returned with `errno==EMFILE`.

FreddieAkeroyd: Do we need to add a separate test for this condition when we check `clientFd` at https://github.com/FreddieAkeroyd/asyn/blob/19db7d33e5576393f99feeb196f88b7d321e5a95/asyn/drvAsynSerial/drvAsynIPServerPort.c#L336  or would this situation also affect `tty->fd` ?

Issue #103, opened 2020-02-13
asynOctetSetOutputEos and asynOctetSetInputEos block IOC init and exit if no device connected


ebernalru:
Hi, 

Setting the OEOS (or IEOS) in the st.cmd file of the testConnect App using an IP that will never connect ends in the IOC blocked a considerable amount of time during init process and at exit.
I am running debian/buster and next configurations:
a) base 3.15.7 + asyn R4-37
b) base 3.15.7 + asyn master (epics-modules/asyn@0679e9f3f757...)
c) base 7.0.3.1(epics-base/epics-base@6feb1c788d...) + asyn master (epics-modules/asyn@0679e9f...)

The modified st.cmd file of the testConnect App looks as follows:

    dbLoadDatabase("../../dbd/testConnect.dbd")
    testConnect_registerRecordDeviceDriver(pdbbase)
    # This is a bogus IP address that will never connect
    drvAsynIPPortConfigure("IPPort", "164.54.160.220:20", 0, 0, 0);
    asynSetTraceIOMask("IPPort",0,0xFF)
    asynOctetSetOutputEos("IPPort",0,"\r")
    iocInit()

No matter what config (a, b or c) i use, the output is very similar, the init takes 2min+10 seconds and the exit between 2 min and 16 min. 

Next output belongs to test case c).
```
vagrant@buster:~/development/asyn/iocBoot/ioctestConnect$ ../../bin/linux-x86_64/testConnect st.cmd 
dbLoadDatabase("../../dbd/testConnect.dbd")
testConnect_registerRecordDeviceDriver(pdbbase)
# This is a bogus IP address that will never connect
drvAsynIPPortConfigure("IPPort", "164.54.160.220:20", 0, 0, 0);
asynSetTraceMask("IPPort",0,0xFF)
asynOctetSetOutputEos("IPPort",0,"\r")
2020/02/13 16:52:04.161 IPPort addr -1 queueRequest priority 3 not lockHolder
2020/02/13 16:54:14.301 asynManager connect queueCallback port:IPPort
2020/02/13 16:54:14.301 IPPort set Eos 1

iocInit()
Starting iocInit
############################################################################
## EPICS R7.0.3.2-DEV
## Rev. R7.0.3.1-78-g6feb1c788d261376d3b1
############################################################################
2020/02/13 16:54:14.302 IPPort -1 autoConnect
2020/02/13 16:54:14.302 Attempting to connect to 164.54.160.220:20  reason:0  fd:-1
[...]

PR #211, opened 2024-12-03
Nonblocking connect using poll

ericonr:
Replacing #210

### DIFF
--- a/asyn/drvAsynSerial/drvAsynIPPort.c
+++ b/asyn/drvAsynSerial/drvAsynIPPort.c
@@ -505,13 +505,56 @@ connectIt(void *drvPvt, asynUser *pasynUser)
             }
         }
 
+    }
+
+#ifdef USE_POLL
+    if (setNonBlock(fd, 1) < 0) {
+        epicsSnprintf(pasynUser->errorMessage,pasynUser->errorMessageSize,
+                               "Can't set %s O_NONBLOCK option: %s",
+                                       tty->IPDeviceName, strerror(SOCKERRNO));
+        epicsSocketDestroy(fd);
+        return asynError;
+    }
+#endif
+
+    if (pasynUser->reason <= 0) {
+
         /*
          * Connect to the remote host
          * If the connect fails, arrange for another DNS lookup in case the
          * problem is just that the device has DHCP'd itself an new number.
          */
         if (tty->socketType != SOCK_DGRAM) {
-            if (connect(fd, &tty->farAddr.oa.sa, (int)tty->farAddrSize) < 0) {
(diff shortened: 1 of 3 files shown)


asyn/drvAsynSerial/drvAsynIPPort.c
```diff
@@ -505,13 +505,56 @@ connectIt(void *drvPvt, asynUser *pasynUser)
...
-            if (connect(fd, &tty->farAddr.oa.sa, (int)tty->farAddrSize) < 0) {
+            int connectResult = connect(fd, &tty->farAddr.oa.sa, (int)tty->farAddrSize);
+        #ifdef USE_POLL
+            if (connectResult < 0 && ((SOCKERRNO == EWOULDBLOCK) || (SOCKERRNO == EINPROGRESS))) {
+                double connectTimeout;
+                int msConnectTimeout;
+                struct pollfd pollfd;
+
+                pasynManager->getAutoConnectTimeout(&connectTimeout);
+                msConnectTimeout = 1000 * connectTimeout;
```
ericonr, line 536: One thing to note, which I don't know if it's relevant, is that any timeouts below 1ms will be rounded down to 0, causing poll to return immediately.

I don't think anyone actually would have such strict timeout requirements, though.

ericonr:
The failure seems unrelated: https://ci.appveyor.com/project/MarkRivers/asyn/builds/51101760/job/iyv53p19fkkop0vd

MarkRivers:
@ericonr I would like to merge this soon.  Can you create a Github issue that describes the problem this PR is designed to fix, and some simple observations that will confirm that it is working?  I think it fixes a problem like this:
```
drvAsynIPPortConfigure("MyPort", "ip_address", 0, 0, 0)
asynOctetSetInputEos("MyPort",0,"\r")
asynOctetSetOutputEos("MyPort",0,"\r")
```
If ip_address is not reachable then the asynOcterSetInputEos() asynOcterSetOutputEos() commands will each hang for a long time because the port is locked waiting for the connect() to time out.  Is this correct?

ericonr:
Hi @MarkRivers 

I have updated the commit message to include a better description of the problem being solved, as well as what was used to confirm it's working. Should I still open an issue, or is the commit description enough?

> If ip_address is not reachable then the asynOcterSetInputEos() asynOcterSetOutputEos() commands will each hang for a long time because the port is locked waiting for the connect() to time out. Is this correct?

That is correct!

chrschroeder:
@MarkRivers @ericonr There is already an open issue for this (https://github.com/epics-modules/asyn/issues/103). I have backported the patches to our local branch and rolled it out to some of our IOCs. Works fine and no problems so far. Thanks for your work to fix this!

ericonr:
@MarkRivers can we merge this and close the issue?

Issue #213, opened 2025-01-14
Update CI build matrix


ralphlange:
# The CI build (using GitHub Actions) lacks maintenance.

I am willing to update things as a one-off project, part of my current CI related PR.

@FreddieAkeroyd @exzombie @MarkRivers @mdavidsaver @keenanlang 
But: You must help and tell me which combinations of build parameters the CI builds should cover.
Obviously, running all possible combinations creates way too many jobs. What we need is a useful, sensible subset.

## Available options (not all shown)

_(Everything italics and in brackets)_ is currently not used

### Platforms (native)
_(Ubuntu 20)_
_(Ubuntu 22)_
Ubuntu 24

Windows-2019
_(Windows-2022)_
_(Windows-2025)_

_(MacOS 13)_
_(MacOS 14)_

### Platforms (cross from Linux to)
RTEMS-4.9
RTEMS-4.10
_(RTEMS-5.1)_
_(Linux@ARM64)_
_(Windows/MinGW)_

### Platforms (using docker)
_(CentOS:8)_
_(RockyLinux:9)_
_(Fedora:33)_
_(Fedora:latest)_

### Sets of dependency versions
_(**dep1**: SNCSEQ 2.2.8, SSCAN 2.11.3, CALC 3.7.3, IPAC 2.15)_
**dep2**: SNCSEQ 2.2.8, SSCAN 2.11.5, CALC 3.7.4, IPAC 2.16
_(**dep-latest**: default branch on all)_

### Configurations
shared (default)
_(debug)_
static (not on MacOS)
_(static-debug (not on MacOS))_

### Compilers (all available in different versions)
GNU (not on MacOS)
_(clang)_
MSVC (only Windows)

### Special cases
_(Only use EPICS build system and libcom (no device support))_
_(Set INSTALL_LOCATION)_

tboegi:
@ralphlange  (hopefully this is useful information)
I am currently using Ubuntu 20:
https://github.com/EuropeanSpallationSource/m-epics-ethercatmc/blob/master/.github/workflows/ci-scripts-build.yml
However, that should be update towards Ubuntu 22.
Or, if nobody else is on Ubuntu 22, to Ubuntu 24.

FreddieAkeroyd:
[...]

PR #214, opened 2025-01-15
Cleanup config and CI

ralphlange:
This PR tries to get the asyn CI into a more reasonable state.
- update ci-scripts-build configuration from Base and other modules
- clean-up the settings mechanism and add settings for synapps releases
- use more comprehensible names for jobs
- fix failing 3.14 builds that were just commented
- make the "current" jobs also use the current commits from dependencies
- add jobs for MacOS and RTEMS-5 (see also #213)
- don't build VXI11 support on RTEMS-5.


ralphlange:
@tboegi has an open PR (#153) to fix the RTEMS-5 issue.

MarkRivers (approved):
This looks good to me.

MarkRivers:
@ralphlange and @hjunkes I would like to revisit part of this PR.  

I don't like the fact that VXI-11 is now disabled for all architectures by default, because DRV_VXI11 is commented out in CONFIG_SITE.  All Linux and Windows users need to edit CONFIG_SITE to enable this, when it was previously enabled by default.

I have 2 questions:
- The new version of asyn/Makefile sets DRV_VXI11=NO for RTEMS:  https://github.com/epics-modules/asyn/blob/87f952ef8ce2dadf7c156fee2a15c801c6796ae1/asyn/Makefile#L247
  Why isn't this sufficient?
- If that is not sufficient then I think it would be preferable to add configure/CONFIG_SITE.RTEMS (or whatever the syntax for that is) with DRV_VX11=NO and let DRV_VX11=YES be the default in CONFIG_SITE.

ralphlange:
Sounds reasonable. I'll look into it.

PR #217, opened 2025-01-29
Queue connect timer on autoconnect enable

FreddieAkeroyd:
Hi @MarkRivers we have an issue with late enabling of autoconnect and I was wondering if this was a valid fix.

 We setup some ports with `drvAsynIPPortConfigure` and autoconnect disabled and then later in SNL we use  `pasynManager->autoConnect`  (like inside the `asynAutoConnect` epics cmd) to enable autoconnect . It seems with this approach that we get an initial connection attempt, but if this fails no further attempts are queued.

Is the correct fix to start a connection timer in the  `autoConnectAsyn` function or is another approach better?

Thanks,

Freddie


exzombie:
I encountered this problem as well, and came to the same solution. Your patch looks good to me.

PR #222, opened 2025-10-15
Fix compile errors in C23 mode

JJL772:
C23 mandates casting between function pointer types, and the compiler can no longer infer arguments when calling functions with empty argument lists.

GCC 15 changes the default C standard to C23, so asyn will only build if `USR_CFLAGS=-std=c17` (for example) is set when running make.


tboegi:
Think that the information from themessage here in the PR 
should be part of the commit message.

Different question:
Does this work with EPICS base 3.15 ?
If not, some inspiration may be taken from here:
https://github.com/epics-modules/motor/blob/master/motorApp/MotorSrc/motor_epics_inc.h
(But there may be other implementations)

JJL772:
@tboegi That's a good question. does 3.15 not have DEVSUPFUN/RECSUPFUN?

appveyor seems to be passing for base 3.14/3.15

PR #226, opened 2026-01-13
Fix compilatition faliure for asynStatus in clang 17

tboegi:
After installing a new clang under MacOS:
```
c++ --version
Apple clang version 17.0.0 (clang-1700.6.3.2)
Target: x86_64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin ----------------
```
the compilation of a motor driver failed:
```
c++ -DUNIX -Ddarwin -O3 -g -Wall -arch x86_64 -fno-common -fPIC -I. -I../O.Common -I. -I. -I.. -I.../epics/modules/motor/include/compiler/clang -I.../epics/modules/motor/include/os/Darwin -I.../epics/modules/motor/include -I.../epics/modules/motor/include -I.../epics/base/../modules/asyn/include -I.../epics/base/include/compiler/clang -I.../epics/base/include/os/Darwin -I.../epics/base/include -c ../smarActMCS2MotorDriver.cpp ../smarActMCS2MotorDriver.cpp:147:10: error: integer value 8 is outside the valid range of
      values [0, 7] for the enumeration type 'asynStatus' [-Wenum-constexpr-conversion]
  147 |     case asynParamWrongType:
      |          ^
.../epics/base/../modules/asyn/include/paramErrors.h:16:32: note:
      expanded from macro 'asynParamWrongType'
   16 | #define asynParamWrongType     (asynStatus)(asynDisabled + 3)
      |                                ^
../smarActMCS2MotorDriver.cpp:149:10: error: integer value 9 is outside the valid range of
      values [0, 7] for the enumeration type 'asynStatus' [-Wenum-constexpr-conversion]
  149 |     case asynParamBadIndex:
      |          ^
.../epics/base/../modules/asyn/include/paramErrors.h:17:32: note:
      expanded from macro 'asynParamBadIndex'
   17 | #define asynParamBadIndex      (asynStatus)(asynDisabled + 4)
      |                                ^
../smarActMCS2MotorDriver.cpp:151:10: error: integer value 10 is outside the valid range of
      values [0, 7] for the enumeration type 'asynStatus' [-Wenum-constexpr-conversion]
  151 |     case asynParamUndefined:
      |          ^
.../epics/base/../modules/asyn/include/paramErrors.h:18:32: note:
      expanded from macro 'asynParamUndefined'
   18 | #define asynParamUndefined     (asynStatus)(asynDisabled + 5)
      |                                ^
3 errors generated.
----------------
```
Solution:
Define all possible values for asynStatus in asynDriver.h

Changes to be committed:
    modified:   asyn/asynDriver/asynDriver.h
    modified:   asyn/asynDriver/asynManager.c
    modified:   asyn/asynPortDriver/paramErrors.h

### DIFF
--- a/asyn/asynDriver/asynDriver.h
+++ b/asyn/asynDriver/asynDriver.h
@@ -46,7 +46,19 @@ extern "C" {
 
 
 typedef enum {
-    asynSuccess,asynTimeout,asynOverflow,asynError,asynDisconnected,asynDisabled
+    asynSuccess,
+    asynTimeout,
+    asynOverflow,
+    asynError,
+    asynDisconnected,
+    asynDisabled,
+    /* Values used for the param interface */
+    asynParamAlreadyExists,
+    asynParamNotFound,
+    asynParamWrongType,
+    asynParamBadIndex,
+    asynParamUndefined,
+    asynParamInvalidList
 }asynStatus;
 
 typedef enum {
--- a/asyn/asynPortDriver/paramErrors.h
+++ b/asyn/asynPortDriver/paramErrors.h
@@ -10,13 +10,4 @@
 
 #include <asynDriver.h>
 
-/* Extend asynManager error list.  We should have a way of knowing what the last error in asyn is */
-#define asynParamAlreadyExists (asynStatus)(asynDisabled + 1)
-#define asynParamNotFound      (asynStatus)(asynDisabled + 2)
-#define asynParamWrongType     (asynStatus)(asynDisabled + 3)
-#define asynParamBadIndex      (asynStatus)(asynDisabled + 4)
-#define asynParamUndefined     (asynStatus)(asynDisabled + 5)
-#define asynParamInvalidList   (asynStatus)(asynDisabled + 6)
-
-
 #endif /* ASYNPORTDRIVERERRORSTATES_H_ */
--- a/asyn/asynDriver/asynManager.c
+++ b/asyn/asynDriver/asynManager.c
@@ -3211,6 +3211,12 @@ static const char *strStatus(asynStatus status)
     case asynError:         return "asynError";
     case asynDisconnected:  return "asynDisconnected";
     case asynDisabled:      return "asynDisabled";
+    case asynParamAlreadyExists: return "asynParamAlreadyExists";
+    case asynParamNotFound: return "asynParamNotFound";
+    case asynParamWrongType: return "asynParamWrongType";
+    case asynParamBadIndex: return "asynParamBadIndex";
+    case asynParamUndefined: return "asynParamUndefined";
+    case asynParamInvalidList: return "asynParamInvalidList";
     }
 
     return "asyn????";


asyn/asynDriver/asynDriver.h
```diff
@@ -46,7 +46,19 @@ extern "C" {
...
+    asynError,
+    asynDisconnected,
+    asynDisabled,
+    /* Extend asynManager error list . */
+    asynParamAlreadyExists,
+    asynParamNotFound,
+    asynParamWrongType,
+    asynParamBadIndex,
+    asynParamUndefined,
+    asynParamInvalidList
```
ericonr, line 61: Have you checked that this doesn't create any new warnings in any asynDriver code, since switch cases might not check for all these enums?

I think the comment is wrong, too. It should mention that the list is extended to cover enums used only by asynPortDriver code.

asyn/asynPortDriver/paramErrors.h
```diff
@@ -10,13 +10,4 @@
 
 #include <asynDriver.h>
 
-/* Extend asynManager error list.  We should have a way of knowing what the last error in asyn is */
-#define asynParamAlreadyExists (asynStatus)(asynDisabled + 1)
-#define asynParamNotFound      (asynStatus)(asynDisabled + 2)
-#define asynParamWrongType     (asynStatus)(asynDisabled + 3)
-#define asynParamBadIndex      (asynStatus)(asynDisabled + 4)
-#define asynParamUndefined     (asynStatus)(asynDisabled + 5)
-#define asynParamInvalidList   (asynStatus)(asynDisabled + 6)
```
ericonr, line 19: I dislike that this commit moves internal errors from asynPortDriver into the asynDriver header. But updating the asynPortDriver functions to return a different kind of error would probably be too complicated...

ericonr (changes requested):
There's a typo in the commit title, "compilatition" should be "compilation".

Do you think the commit message should include the whole compiler output? I think it could be resumed into a simpler and shorter text mentioning `-Wenum-constexpr-conversion` and the fact that asynPortDriver was extending an enum by defining additional constants (which is not possible in C++). The compiler version (clang-17) should also be enough.

The solution should mention that we are changing these "internal" enums into public ones, and address any (if there are) concerns over exposing that in the public interface. Is there another way to solve this? Should they be named something like  `asyn_internal__asynParamAlreadyExists` or similar so that `paramErrors.h` can do `#define asynParamAlreadyExists asyn_internal__asynParamAlreadyExists`?

Finally, I don't think the commit should mention which files it touches; that information can be obtained with `git log --name-status`, after all.

mdavidsaver:
> changing these "internal" enums into public ones

`paramErrors.h` is being installed.  Are these "extended" error codes intended to be internal?

https://github.com/epics-modules/asyn/blob/e2a281e22b4d84ba51e56428959c2dfa506a8312/asyn/Makefile#L135

The intent is not clear to me from looking at the history: 55c33f340ffccec6fa13891b54593cd37ac7f826 a9165a8dbc91f615185914a8a6d7991546c0eeae

asyn/asynDriver/asynDriver.h
mdavidsaver: > Have you checked that this doesn't create any new warnings in any asynDriver code, since switch cases might not check for all these enums?

I see at least once `switch` which should be updated.

https://github.com/epics-modules/asyn/blob/e2a281e22b4d84ba51e56428959c2dfa506a8312/asyn/asynDriver/asynManager.c#L3205-L3216

Glancing through the output of `git grep switch`, there are likely one or two others.

tboegi:
@ericonr @mdavidsaver Thanks for the reviews. A new patch should address your comments

asyn/asynDriver/asynDriver.h
```diff
@@ -46,7 +46,19 @@ extern "C" {
...
 
 typedef enum {
-    asynSuccess,asynTimeout,asynOverflow,asynError,asynDisconnected,asynDisabled
+    asynSuccess,
+    asynTimeout,
+    asynOverflow,
+    asynError,
+    asynDisconnected,
+    asynDisabled,
+    /* Values used for the param interface */
```
ericonr, line 55:
```suggestion
    /* Values used for the asynPortDriver param interface */
```

ericonr (commented):
I assumed the header was internal. If it's already installed, it doesn't seem like such a big deal to have everything in the same bag of enums.

PR #229, opened 2026-02-25
Add a missing exception string

exzombie:
This fixes the mismatch in the length of the `asynException` enum and  the `ASYN_EXCEPTION_STRINGS` macro. It also fixes the length check in `asynRecord`.

### DIFF
--- a/asyn/asynDriver/asynDriver.h
+++ b/asyn/asynDriver/asynDriver.h
@@ -59,7 +59,8 @@ typedef enum {
 #define ASYN_EXCEPTION_STRINGS                                                          \
     "asynExceptionConnect",   "asynExceptionEnable",      "asynExceptionAutoConnect",   \
     "asynExceptionTraceMask", "asynExceptionTraceIOMask", "asynExceptionTraceInfoMask", \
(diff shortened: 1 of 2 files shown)


asyn/asynDriver/asynDriver.h
```diff
@@ -59,7 +59,7 @@ typedef enum {
 #define ASYN_EXCEPTION_STRINGS                                                          \
     "asynExceptionConnect",   "asynExceptionEnable",      "asynExceptionAutoConnect",   \
     "asynExceptionTraceMask", "asynExceptionTraceIOMask", "asynExceptionTraceInfoMask", \
-    "asynExceptionTraceFile", "asynExceptionTraceIOTruncateSize"
+    "asynExceptionTraceFile", "asynExceptionTraceIOTruncateSize", "asynExceptionShutdown"
```
ericonr: My only nit is that reviewing would be slightly easier if it followed the same organization as the enum definition above, so `"asynExceptionShutdown"` would be in its own line.

exzombie: Agreed.

ericonr (approved):

